PR #8, opened 2024-01-10
i18n(sr): Update translations

git-f0x:
Unrelated to the pull request but didn't want to create a separate issue over something that isn't that important:

Perhaps the nav bar selection graphic should either be a bit longer or slightly thinner, since it's currently a kind of elongated squircle that doesn't match the pill shape present in cosmic-settings or cosmic-edit (though it had a pill shape before the `reduce width` commit, so might be better to just make it thinner). 

And perhaps the name of the Home directory in the nav bar should be just Home instead of the user name, to match other file managers.

If this was already planned or is intended, sorry for making you read all of this. Cosmic Files is starting to look pretty nice!

Edit: Just saw that the old Figma design draft has all of this, so maybe all this text was unnecessary.

### DIFF
--- a/i18n/sr-Cyrl/cosmic_files.ftl
+++ b/i18n/sr-Cyrl/cosmic_files.ftl
@@ -1,5 +1,6 @@
 empty-folder = Празна фасцикла
 empty-folder-hidden = Празна фасцикла (има скривене ставке)
+filesystem = Систем датотека
 trash = Отпад
 
 # Context Pages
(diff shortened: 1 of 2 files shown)


jackpot51:
Thanks, yes it is likely we will substitute "Home" (or a translation) for the name of the home folder. I'm not sure what caused the shape of the nav bar selector to change but I will look into it

git-f0x:
Not sure if you'll see this but I just saw that `cosmic-settings` now also has a similar shape of the selector (though the edges are rounder than in `cosmic-files`) but `cosmic-edit` doesn't. Not sure when it changed.
Screenshot_2024-01-10_22-28-33

jackpot51:
Probably an update to libcosmic caused this.

PR #16, opened 2024-01-15
Add capability to access remote filesystems

jackpot51:


jackpot51:
Mostly resolved by GVFs support

Issue #39, opened 2024-02-13
No fast way to close "Complete" notification in right side drawer thing


leviport:
When I delete a file, the right hand drawer opens with the notification that the operation has been completed. This makes the rest of the file browser insensitive, so I have to close it before I can continue with whatever I was doing. Clicking files does not close it, and it blocks my click. Hitting Esc also does not close it, so I have to use my mouse.

maria-komarova:
I don't see that happening now so maybe not relevant anymore?

leviport:
I also don't see the Operations pop-over appear anymore. I'm not sure if that's a whole new bug though. Should it be appearing?

Also, if I open the Operations drawer with Edit > Operations, I can now close it with Esc, but I think clicking back on the parent window should probably close it also. If it's supposed to, I can open a bug report.

maria-komarova:
Context drawer is generally only closed by clicking on Close now. We can think about closing it on clicking outside in certain cases but it is not a menu.
I'm not sure about Operations context drawer. Was it appearing when you Deleted an item before? I don't think that should be the case.

PR #49, opened 2024-02-22
Close context pages and tab menus on Escape

joshuamegnauth54:
Fixes (or at least improves): #39 

This doesn't close the `MenuBar` if it's opened yet because I couldn't figure out a good way to do it.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -575,6 +575,22 @@ impl Application for App {
         Command::none()
     }
 
+    fn on_escape(&mut self) -> Command<Self::Message> {
+        let entity = self.tab_model.active();
+
+        // Close menus and context panes in order per message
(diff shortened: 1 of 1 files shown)


src/app.rs
```diff
@@ -617,6 +624,19 @@ impl Application for App {
                     return self.update_config();
                 }
             }
+            Message::ContextCloseAny => {
+                let entity = self.tab_model.active();
+
+                // Close menus and context panes in order per message
```
jackpot51, line 630: Rather than do a Message, I would recommend to move this code to the on_escape function

joshuamegnauth54: Alright. I refactored it.

jackpot51 (approved):

PR #55, opened 2024-02-28
Filtering of files in search

leb-kuchen:
This kinda works like in nautilis, so only path forward.

### DIFF
--- a/src/tab.rs
+++ b/src/tab.rs
@@ -1033,6 +1033,9 @@ impl Tab {
                     hidden += 1;
                     continue;
                 }
+                if self.filter_search(item) {
+                    continue;
+                }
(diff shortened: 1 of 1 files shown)


jackpot51:
Are there missing changes that go with this?

leb-kuchen:
You can only filter the items in the list, else you would have to update it each time the search changes.

jackpot51:
There are some issues with this. When it filters it shows the empty folder icon and text and doesn't indicate that search is hiding items. That location editing is not meant to be used for search, it should offer to autocomplete the path instead.

leb-kuchen:
1 Search and 1 Autocomplete?

PR #63, opened 2024-03-01
Visible columns

YkdWaWEzVmphR1Z1:


src/tab.rs
```diff
@@ -1689,46 +1738,71 @@ impl Tab {
...
-                    .style(button_style(item.selected, true))
-                    .on_press(Message::Click(Some(i)));
+                let columns_content = visible_columns.iter().filter(|col| col.active).map(|col| {
+                    widget::text(column_text(col.heading))
+                        .width(get_column_width(col.heading))
+                        .into()
+                });
+                columns.extend(columns_content);
+
+                // let row = if condensed {
```
jackpot51, line 1762: The condensed view is important for being able to adapt to small window sizes. Why is it removed here?

leb-kuchen:
I see, wasn't there when i started, but is indeed important.

jackpot51:
Thanks for that! I am talking with the team as we haven't made our designs public yet but we need to match further changes to them. Hopefully we can share something early next week.

PR #75, opened 2024-03-15
Sorter choice config + sort grid view

joshuamegnauth54:
I've implemented an option to set a default sorting scheme for new tabs. This is related to but doesn't close [#149](https://github.com/pop-os/cosmic-edit/issues/149) on COSMIC Edit.

This PR also adds a feature to sort the grid view by reusing the code to sort the list view.

Finally, I've also added context menu buttons to sort the views without going to settings - this is more useful for grid view. The new context menu items are similar to what is offered in Thunar, Windows Explorer, and likely others.


jackpot51 (approved):
This may be adjusted later based on UX feedback, but it is working.

Issue #48, opened 2024-02-19
No images renderd for icons


WhyNotHugo:
First of all, I'll clarify that I'm not using cosmic desktop; I'm trying to use cosmic-files under swaywm.

I'm trying out `cosmic-files`, and files don't render icons. The console output reads lots of lines like:

```
[2024-02-19T18:47:26Z WARN  cosmic_files::mime_icon] failed to get icon for MimeIconKey { path: "/home/hugo/downloads/Untitled.svg", size: 32 }: (icon not found, (Error type: IoError)
```

A quick glace at the code indicates that this is in `src/mime_icon.rs`, `MimeIconCache::get` calls `systemicons::get_icon`. I looked at its code and it seems to use Gtk under the hood. I'm using [`nemo`](https://github.com/linuxmint/nemo) (gtk3) as a file manager and icons for it render fine, so I'm fairly certain that a theme is properly installed and configured.

WhyNotHugo:
Huh, running the example from https://github.com/uriegel/systemicons/ fails, so maybe the bug is not specific to this crate?

Any special dependency that I should keep in mind in this context?

WhyNotHugo:
> Huh, running the example from https://github.com/uriegel/systemicons/ fails, so maybe the bug is not specific to this crate?

The example's failure seem unrelated, since it's actually SEGFAULT'ing.

joshuamegnauth54:
I'm able to replicate this on my Sway machine too. COSMIC Files is no longer using the crate above but instead using features exposed by `libcosmic` directly. `libcosmic` is using [freedesktop-icons](https://github.com/oknozor/freedesktop-icons) instead of the (likely unmaintained?) crate above.

In any case, I tested all of this with a quick program that you can try too:

```rust
use std::{fs, path::Path};

use cosmic::widget::icon;
use xdg_mime::SharedMimeInfo;

fn main() {
    let arg = std::env::args().nth(1).unwrap();
    let path = Path::new(&arg);
    let mime_info = SharedMimeInfo::new();

    let mimes = if path.is_file() {
        mime_info.get_mime_types_from_file_name(&arg)
    } else {
        let mut mimes = vec![];
        for entry in fs::read_dir(path).unwrap() {
            let entry = entry.unwrap().path();
            let path = entry.to_str().unwrap();
            mimes.extend(mime_info.get_mime_types_from_file_name(path));
        }
        mimes
    };

    for mime in mimes {
        println!("{mime}");
[...]

PR #102, opened 2024-04-04
Icon theme fallback for missing themes

joshuamegnauth54:
Fixes: #48

This PR implements a last resort fallback to the default GTK icon theme if COSMIC's theme is missing.

The user is warned if COSMIC's themes aren't available since a correct installation should have them much like GTK should always have Adwaita.


TheESN:
Also some icon themes specify that they rely on other icon theme as fallback when something is missing. COSMIC doesn't implement this behavior at this moment, so it would be neat to have this

mmstick:
This should be implemented in libcosmic instead. Avoid the duplication across projects.

PR #103, opened 2024-04-05
Dnd

wash2:


wash2:
Ok this is working well now, but it still needs support for receiving dragged items in the nav bar.

jackpot51:
This works pretty well, but it appears to interfere with selection if an item is already selected. I think drop should only start if an item is selected and the drag started from the item.

jackpot51:
To reproduce the issue, select an item, then drag it, then attempt to select other items via dragging from outside of that item.

jackpot51:
Additionally, it looks like drag select does not allow releasing the mouse button on top of an item

edfloreshz:
If you drag a folder onto itself the folder gets deleted, no apparent way to recover it.

Try to reproduce with unimportant folders.

wash2:
> Should I open an issue for this?

Yes, I hope you haven't lost anything as a result of this :(

edfloreshz:
Luckily only my `Videos` folder, which iirc had nothing important.

lilyinstarlight:
It looks like this broke test `tab::tests::tab_click_double_opens_folder`, and it's still broken on main branch

If no one looks at it before then, I might look at it in the next few days (probably Monday though)

```
running 11 tests
test tab::tests::scan_path_empty_dir_returns_empty_vec ... ok
test tab::tests::tab_empty_history_does_nothing_on_prev_next ... ok
test tab::tests::tab_locationup_moves_up_hierarchy ... ok
test tab::tests::scan_path_returns_empty_vec_for_invalid_path ... ok
test tab::tests::tab_location_changes_location ... ok
test tab::tests::tab_goprev_moves_backward_in_history ... ok
test tab::tests::tab_gonext_moves_forward_in_history ... ok
test tab::tests::tab_click_ctrl_selects_multiple ... ok
test tab::tests::tab_click_single_selects_item ... ok
test tab::tests::scan_path_succeeds_on_valid_path ... ok
test tab::tests::tab_click_double_opens_folder ... FAILED

failures:

---- tab::tests::tab_click_double_opens_folder stdout ----
thread 'tab::tests::tab_click_double_opens_folder' panicked at src/app.rs:2196:9:
assertion `left == right` failed: Tab's path is /tmp/.tmpTYwIEx instead of being updated to /tmp/.tmpTYwIEx/sETYe
  left: "/tmp/.tmpTYwIEx/sETYe"
 right: "/tmp/.tmpTYwIEx"
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    tab::tests::tab_click_double_opens_folder

test result: FAILED. 10 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.02s
```

Issue #151, opened 2024-05-09
Navigating with the arrow keys after clicking on a file renders the first arrow key "empty" before the navigation works


onlyreportingissues:
Not sure how to explain it, but the first arrow key doesn't navigate at all, only after the first one the navigation works.

Issue #114, opened 2024-04-12
Select multiple items using Shift and click


maria-komarova:
Click on one item, hold Shift and then click on another item should select the items in between the two clicked items.

MohammadShabaniSBU:
Hi, I am trying to implement this feature, but I noticed that when a click event is triggered, the match arm for Message::Click executes twice. I want to know if it is a bug or something that you implemented.

jasonrhansen:
@MohammadShabaniSBU I noticed the same issue where the Click event would fire twice. I fixed it in my first commit of #193.

PR #193, opened 2024-05-27
Selection enhancements and bug fixes

jasonrhansen:
- Fix for extra click event. When clicking a file with the mouse, two `Click` events would get fired instead of one.
- Better ctrl + click handling. Clicking on an already selected item will deselect it on `ClickRelease`.
- Add support for drag-selecting with shift of ctrl pressed. This will now invert the selection of any items within the drag rectangle.
- Implement shift+click selection. #114
- Fix for arrow key navigation/selection after an item is clicked with the mouse. #151
- Fix a bug where drag-and-drop would sometimes get initiated inadvertently when there were files selected but you clicked and dragged in the empty area below the files. This should instead start a new drag selection.

### DIFF
--- a/src/mouse_area.rs
+++ b/src/mouse_area.rs
@@ -474,9 +474,7 @@ fn update<Message: Clone>(
             state.drag_initiated = cursor.position();
         }
 
-        if let Some(message) = widget.on_press.as_ref() {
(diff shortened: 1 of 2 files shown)


jackpot51:
Great work! I will review this soon.

jackpot51 (approved):

PR #201, opened 2024-05-29
Update com.system76.CosmicFiles.desktop

VandaLHJ:
Not sure if it's ok, but i added PL translation bits, so it looks more native for PL users if locale is set.


jackpot51:
This is good, thanks!

jackpot51 (approved):

Issue #262, opened 2024-06-21
Allow using context menu on breadcrumbs


jacobgkau:
It would be nice to be able to right-click a breadcrumb and open it in a new tab (or middle-click the breadcrumb, if https://github.com/pop-os/cosmic-files/issues/261 is implemented).

Currently, if I'm deep in a directory structure (levels away from my favorites) and want to open a sibling folder in another tab, there's no quick & easy way to do that (the closest thing is to navigate back up a level in the current tab, open the sibling folder in a new tab, then re-open the folder I was already in before in the current tab).

Issue #261, opened 2024-06-21
Middle-click to open in background


jacobgkau:
Similar to Nautilus and Dolphin, it would be nice to be able to middle-click folders (in the main view, sidebar, and breadcrumbs) to be able to open them in a new tab, instead of having to right-click and use the context menu.

jacobgkau:
This seems to be working now. This does focus the new tab immediately, which makes it take longer to e.g. open a row of adjacent folders in separate tabs. Nautilus and Dolphin both open tabs in the background (similarly to web browsers opening links in the background when middle-clicking). @pop-os/ux Can you confirm if focusing the new tab is the intended behavior?

maria-komarova:
After some internal discussion and more thinking, let's stick with what is familiar to people from Nautilus.

WatchMkr:
Currently opens and focuses tab. Open in background instead.

LeoBeliik:
> Open in background instead.

I would actually love a toggle in the config for that, I understand everyone's workflow may differ, for me if I'm opening a new tab is because I want to do something in that folder so having it on focus would be desirable.
Furthermore, Dolphin (I don't know if it's the case in Nautilus), let's you auto focus to the new tab by using Shift + Middle click, right now Cosmic Files doesn't have this option (it just select folders like if you were using shift right click).
The desirable feature here would be to have a toggle in the app config to change if you want new tabs to auto focus or not, and shift + middle click would do the opposite.

Thanks!

PR #266, opened 2024-06-30
Open in new tab/window

jasonrhansen:
* Add "Open in new tab" and "Open in new window" to context menu.
* Add bread crumb context menu with "Open in new tab", "Open in new window", and "Properties".
* Open files with middle click.
* Open directories in a new tab with middle click.
  * Works in the main view.
  * Works for the bread crumbs,
  * Works in nav bar.
* Add key bindings to open in new tab (Ctrl + Enter) and open in new window (Shift + Enter).

Implements #261.
Implements #262.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1105,7 +1105,7 @@ dependencies = [
 [[package]]
 name = "cosmic-config"
 version = "0.1.0"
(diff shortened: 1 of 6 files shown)


jackpot51 (approved):
Excellent work, thank you!

Issue #18, opened 2024-01-15
[Feature Request] cosmic-file zoom


gantryyork:
It would be nice if cosmic-file would allow for zooming 

- cntr with scroll wheel
- cntr + and cntr -

l-const:
This should be closed now after the merge of https://github.com/pop-os/cosmic-files/pull/455

PR #271, opened 2024-07-06
Add 'Ctrl' + '+' as a zoom-in shortcut

Kartonrealista:
It seems the zoom in shortcut was set to 'Ctrl' + '='  to make it work better with 70% and smaller keyboards, but 'Ctrl' + '+' wasn't added.


Kartonrealista:
#18

jackpot51 (approved):

Issue #258, opened 2024-06-20
panics when sorting files with names as big numbers


hey-adora:
Panics on trying to sort files at `tab.rs` in `scan_path` function at line 315

```
thread 'tokio-runtime-worker' panicked at /home/hey/.cargo/registry/src/index.crates.io-6f17d22bba15001f/lexical-sort-0.3.1/src/cmp.rs:106:21:
attempt to multiply with overflow
stack backtrace:
   0: rust_begin_unwind
             at /rustc/129f3b9964af4d4a709d1383930ade12dfe7c081/library/std/src/panicking.rs:652:5
   1: core::panicking::panic_fmt
             at /rustc/129f3b9964af4d4a709d1383930ade12dfe7c081/library/core/src/panicking.rs:72:14
   2: core::panicking::panic_const::panic_const_mul_overflow
             at /rustc/129f3b9964af4d4a709d1383930ade12dfe7c081/library/core/src/panicking.rs:180:21
   3: lexical_sort::cmp::natural_lexical_cmp
   4: cosmic_files::tab::scan_path::{{closure}}
             at ./src/tab.rs:315:14
   5: alloc::slice::<impl [T]>::sort_by::{{closure}}
             at /rustc/129f3b9964af4d4a709d1383930ade12dfe7c081/library/alloc/src/slice.rs:267:34
   6: core::slice::sort::merge
             at /rustc/129f3b9964af4d4a709d1383930ade12dfe7c081/library/core/src/slice/sort.rs:963:28
   7: core::slice::sort::merge_sort
             at /rustc/129f3b9964af4d4a709d1383930ade12dfe7c081/library/core/src/slice/sort.rs:1102:17
   8: alloc::slice::stable_sort
             at /rustc/129f3b9964af4d4a709d1383930ade12dfe7c081/library/alloc/src/slice.rs:887:5
   9: alloc::slice::<impl [T]>::sort_by
   ```

![image](https://github.com/pop-os/cosmic-files/assets/146812294/681f0c5a-a90e-4a16-817c-0194eaa96e97)

looks like the bug is in the lexical-sort repo
but it seems the fix is not being merged for a year now
https://github.com/Aloso/lexical-sort/pull/11

joshuamegnauth54:
[...]

PR #272, opened 2024-07-08
fix: Panic on long numerical file names (replace lexical_sort)

joshuamegnauth54:
Closes: #258 

`lexical_sort` is unmaintained as of two months ago. There is an old PR that fixes the issue mentioned by #258, but even if that patch is merged the crate is still abandoned.

 `lexical_sort`'s author recommends using the `icu` crates which are maintained by the Unicode Consortium. I think `icu` can also be used to localize date strings which is another open issue.


jackpot51:
lexical-sort is also being used in cosmic-edit, cosmic-store, and cosmic-term.

jackpot51 (approved):

PR #276, opened 2024-07-11
Design matching

git-f0x:
Made the List view sorting headings fixed size to match the behavior of most file managers (and so that they don't become too large). The designs use 24.0 (space_m), but the text doesn't fit at that size, so 32.0 was used.

Also added some horizontal padding to the breadcrumb bar to better match the designs.

With changes:
screenshot-2024-07-11-22-15-54
Design:
Home - list

Unrelated, but with the latest commits, it seems a 50% list size displays properly, so maybe the bottom end of the list view size range can be lowered. :)

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -801,7 +801,7 @@ impl App {
                     widget::settings::item::builder(fl!("icon-size-list"))
                         .description(format!("{}%", list))
                         .control(
-                            widget::slider(75..=500, list, move |list| {
+                            widget::slider(50..=500, list, move |list| {
                                 Message::TabConfig(TabConfig {
                                     icon_sizes: IconSizes {
                                         list: NonZeroU16::new(list).unwrap(),
--- a/src/app.rs
+++ b/src/app.rs
@@ -1,14 +1,11 @@
 // Copyright 2023 System76 <info@system76.com>
 // SPDX-License-Identifier: GPL-3.0-only
 
-use cosmic::iced::clipboard::dnd::DndAction;
-use cosmic::widget::dnd_destination::DragId;
-use cosmic::widget::menu::action::MenuAction;
-use cosmic::widget::menu::key_bind::KeyBind;
 use cosmic::{
     app::{message, Command, Core},
     cosmic_config, cosmic_theme, executor,
     iced::{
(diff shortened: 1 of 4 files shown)


jackpot51:
If you want to set the lower bound on list size to 50% I will review that too, thanks!

git-f0x:
Done. Now if only I could figure out how to add the accent-colored line under the breadcrumb... :)

Unrelated, but possibly a small issue: if the default view is different from the current view (e.g. default is list, current is grid), changing anything in settings (e.g. show hidden files) makes Files switch back to the default view.
Also is it planned to have a toggle button somewhere for switching views (similar to nautilus), so that the default view options in settings aren't necessary?

jackpot51:
The default text size of 14 pixels has a line height is 20 pixels which is not going to fit in a 16 pixels tall element. Either the minimum list size needs to be increased or the text size must shrink when it is this small.

git-f0x:
The list and font size look good, so is it possible to somehow increase the vertical space for the text, without changing list or font size (maybe some tweaks to padding)?
screenshot-2024-07-12-14-23-37
Nautilus is even smaller at 50%, but the text does shrink at that size.

jackpot51:
I think the padding can be reduced by one level when the icon size is less than 24.

git-f0x:
Would something like this be acceptable? `.padding(if icon_size < 24 { 7 } else { space_xxs })`
Or just to use 7 as a value since the difference is barely perceptible at larger sizes?

git-f0x:
This is pretty much ready for review. :)
screenshot-2024-07-13-16-03-51
screenshot-2024-07-13-16-04-54

I commented the lines preceded by HACK, since removing them fixes the scrollbar after the changes (it was always displayed). If they're there for a different reason, or if I should fully remove them, let me know.

TheESN:
Looks really nice. The only thing that is remaining is making that line coloured with accent colour

git-f0x:
I haven't been able to find any examples of coloring that component, so I'm either not using the correct thing, or it isn't implemented yet. But will probably find out when it's reviewed.

jackpot51 (approved):

Issue #353, opened 2024-08-11
[Feature request] Bookmarks


francesco-gaglione:
In Gnome file you can mark a file or directory as preferred. This is then shown in the favourites tab. 
Currently, I'm trying to implement this in my local environment. 
Does this need a design discussion first?

![image](https://github.com/user-attachments/assets/e143ba6f-380c-4554-8521-1afdc809c955)
![image](https://github.com/user-attachments/assets/23c40099-f490-403d-9e86-42bc7ef443a6)

KodeBarista:
Expected Behavior:
- Users should be able to star files and folders from the table column or context menu.
- A dedicated Starred section should be available (preferably at the top).
- Starred items should be collected in one place for quick access.
- Users should be able to easily add and remove items from the Starred section.

<img width="1462" height="592" alt="Image" src="https://github.com/user-attachments/assets/2e198fe4-f08a-414d-80f1-118d7109428c" />

MikhailValerie:
I have created a Pull Request for implementing "Bookmarks" via the Freedesktop standard user-places file, found here: https://github.com/pop-os/cosmic-files/pull/1982 , with both sidebar-pinned Location and list view column.
I'm keen to see what you think of it, if you don't mind checking it out and giving feedback.

PR #354, opened 2024-08-11
Bookmarks

francesco-gaglione:
Pr linked with #353 

Working on this while waiting for the design discussion


francesco-gaglione:
The implementation in this pull request is a working implementation. I know that this functionality needs to be revised by the graphics team. In the meantime, I have implemented this functionality by modifying the UI/UX as little as possible. 
Initially I wanted to put icons in the list view as in the gnome file, but this was not possible due to a bug in the row button. So I moved the addition/removal of the bookmark to the context menu (this way the UI is not changed). 
The current implementation will probably not be the final one, it can be extended further (especially in terms of the UI) but it is already working and may be useful for alpha 2.

francesco-gaglione:
We should be compliant with freedesktop standard: [freedesktop](https://www.freedesktop.org/wiki/Specifications/desktop-bookmark-spec/). Waiting for https://github.com/pop-os/recently-used-xbel/pull/1

Issue #122, opened 2024-04-18
[Feature Request] Compress/Extract Options


nathansgithub:
One feature I really like in Nautilus is the right-click archive options, and I would love if you could replicate that here.

This is the "Extract Here" and "Extract to" options on various formats of zipped/archive file, and the "Compress" option on any one or more selected files.

nathansgithub:
If you all are willing to merge this, I might take a stab at coding these.

WatchMkr:
@nathansgithub this is a feature we'd merge if implemented.

WatchMkr:
Secondary click options:

Compress
Extract
Extract to...

MVP for compress > make a .zip archive in the current directory. MVP+1 password protected zip archives with dialog for choice and name.

iamkartiknayak:
This feature is needed ASAP. I'm tired of launching terminal every-time to extract a compressed file.

nathansgithub:
I made a PR for an MVP to handle extracting .zip files to the current directory: https://github.com/pop-os/cosmic-files/pull/370

Adding dialogues for the other pieces will be a little more involved, and I'm not sure what the preferred way to handle other archive types would be, but I think this piece is useful by itself.

PR #370, opened 2024-08-12
Enable extracting zip files to current folder

nathansgithub:
Linked to https://github.com/pop-os/cosmic-files/issues/122

This adds an "Extract here" menu option for selected .zip files.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,7 @@ rayon = "1"
 regex = "1"
 serde = { version = "1", features = ["serde_derive"] }
 shlex = { version = "1.3" }
+tar = "0.4.41"
 tokio = { version = "1", features = ["sync"] }
 trash = { git = "https://github.com/jackpot51/trash-rs.git", branch = "delete-info" }
 xdg = { version = "2.5.2", optional = true }
@@ -46,6 +47,7 @@ i18n-embed = { version = "0.14", features = [
 i18n-embed-fl = "0.7"
 rust-embed = "8"
 slotmap = "1.0.7"
+zip = "2.1.6"
 
 [dependencies.libcosmic]
 git = "https://github.com/pop-os/libcosmic.git"
--- a/src/operation.rs
+++ b/src/operation.rs
@@ -1,8 +1,11 @@
 use cosmic::iced::futures::{channel::mpsc::Sender, executor, SinkExt};
+use mime_guess::MimeGuess;
 use std::{
     borrow::Cow,
     fs,
+    io::{self, Error},
     path::{Path, PathBuf},
+    process,
     sync::{
         atomic::{self, AtomicU64},
         Arc,
@@ -138,6 +141,11 @@ pub enum Operation {
     },
     /// Empty the trash
     EmptyTrash,
+    /// Uncompress files
+    Extract {
+        paths: Vec<PathBuf>,
+        to: PathBuf,
+    },
     /// Move items
     Move {
         paths: Vec<PathBuf>,
@@ -246,6 +254,12 @@ impl Operation {
                 to = fl!("trash")
             ),
             Self::EmptyTrash => fl!("emptying-trash"),
+            Self::Extract { paths, to } => fl!(
+                "extracting",
+                items = paths.len(),
+                from = paths_parent_name(paths),
+                to = file_name(to)
+            ),
             Self::Move { paths, to } => fl!(
                 "moving",
                 items = paths.len(),
@@ -284,6 +298,12 @@ impl Operation {
                 to = fl!("trash")
             ),
             Self::EmptyTrash => fl!("emptied-trash"),
+            Self::Extract { paths, to } => fl!(
+                "extracted",
+                items = paths.len(),
+                from = paths_parent_name(paths),
+                to = file_name(to)
+            ),
             Self::Move { paths, to } => fl!(
                 "moved",
                 items = paths.len(),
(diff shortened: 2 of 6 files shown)


src/operation.rs
```diff
@@ -473,6 +491,22 @@ impl Operation {
                     .send(Message::PendingProgress(id, 100.0))
                     .await;
             }
+            Self::Extract { paths, to } => {
+                for path in paths {
+                    let to = path.file_stem().map(|stem| to.join(stem)).unwrap();
+                    tokio::task::spawn_blocking(move || {
+                        process::Command::new("unzip")
```
jackpot51, line 498: An implementation of this should not call out to an external program

jackpot51 (changes requested):

src/operation.rs
nathansgithub: Thank you for looking at this. Do you mean I should import a library to handle extracting files instead of using unzip? Are there any in particular you'd trust to use for this?

nathansgithub: I'm updating it to use the `zip` crate, which looks good for this.

jackpot51: That is fine. More archive types should be supported if possible.

nathansgithub: Definitely want to see more types, and I might keep going on this, but I wanted to keep the scope here small but usable at first. 

Let me know how this looks. I changed it to use the `zip` crate, and also changed it to create duplicate output folders if you extract multiple times, like the copy function.

src/operation.rs
```diff
@@ -473,6 +492,68 @@ impl Operation {
...
+                            let mut extensionless_path = path.to_owned();
+                            extensionless_path.set_extension("");
+                            to = copy_unique_path(&extensionless_path, &to.parent().unwrap());
+                        };
+
+                        fs::create_dir_all(&to)?;
+
+                        let mut archive = fs::File::open(path)
+                            .map(|file| zip::ZipArchive::new(file))?
+                            .expect("Expected valid file");
```
michalfita, line 512: That's going to panic if archive file not valid.

Cargo.toml
```diff
@@ -46,6 +46,7 @@ i18n-embed = { version = "0.14", features = [
 i18n-embed-fl = "0.7"
 rust-embed = "8"
 slotmap = "1.0.7"
+zip = "1.1.4"
```
michalfita, line 49: Why so old `zip` crate?  MSRV?

src/operation.rs
```diff
@@ -473,6 +492,68 @@ impl Operation {
...
+
+                        if to.exists() {
+                            let mut extensionless_path = path.to_owned();
+                            extensionless_path.set_extension("");
+                            to = copy_unique_path(&extensionless_path, &to.parent().unwrap());
+                        };
+
+                        fs::create_dir_all(&to)?;
+
+                        let mut archive = fs::File::open(path)
```
michalfita, line 510: It would make sense if the extractor is in the separate function, the level of nesting makes this hard to review even on a large screen.

src/operation.rs
```diff
@@ -473,6 +492,68 @@ impl Operation {
...
+                for path in paths {
+                    let to = to.to_owned();
+
+                    tokio::task::spawn_blocking(move || -> Result<(), Error> {
+                        let mut to = path.file_stem().map(|stem| to.join(stem)).unwrap();
+
+                        if to.exists() {
+                            let mut extensionless_path = path.to_owned();
+                            extensionless_path.set_extension("");
+                            to = copy_unique_path(&extensionless_path, &to.parent().unwrap());
```
michalfita, line 505: `.unwrap()`s in production code?

src/operation.rs
```diff
@@ -473,6 +492,68 @@ impl Operation {
...
+                                        }
+                                    }
+
+                                    if zipped_file.is_dir() {
+                                        fs::create_dir_all(&output_dir)
+                                    } else {
+                                        fs::File::create(&output_dir).and_then(|mut output_file| {
+                                            io::copy(&mut zipped_file, &mut output_file).map(|_| ())
+                                        })
+                                    }?;
```
michalfita, line 533: Duplicated logic in two different implementations??

michalfita (commented):
Did some fishing in that code:

src/operation.rs
leb-kuchen, line 510: This may be very slow, if ZipArchive does not have its own buffering, so consider using BufReader.

src/operation.rs
```diff
@@ -473,6 +492,68 @@ impl Operation {
                     .send(Message::PendingProgress(id, 100.0))
                     .await;
             }
+            Self::Extract { paths, to } => {
```
leb-kuchen, line 495: ZipArchive also has the method extract,which also handles symlinks.

michalfita: `ZipArchive::`[`extract()`](https://docs.rs/zip/1.1.4/zip/read/struct.ZipArchive.html#method.extract) what disappeared from my review. Especially as current code in this PR does directory extraction twice.

nathansgithub (commented):
@michalfita @leb-kuchen Thank you for the feedback. I somehow missed the extract function. That simplified the code a lot.

@jackpot51 I'd have to do more research into archive crates, but in the meantime, I added [MIME type matching](https://github.com/pop-os/cosmic-files/pull/370/files#diff-a7301f9a3dcd2e25ff77be69aa773585260086ada5c48a4b87c512aca41b1ba6R512) to make it easy to plug in other extraction libraries.

src/operation.rs
```diff
@@ -473,6 +493,48 @@ impl Operation {
...
+                            if new_dir.exists() {
+                                let mut extensionless_path = path.to_owned();
+                                extensionless_path.set_extension("");
+                                if let Some(new_dir_parent) = new_dir.parent() {
+                                    new_dir = copy_unique_path(&extensionless_path, new_dir_parent);
+                                }
+                            }
+
+                            if let Some(mime) = mime_guess::from_path(&path).first() {
+                                match mime.essence_str() {
```
michalfita, line 512: The `zip` crate supports more: LZMA, BZIP2, ZSTD, but looks like w/o handling `tar`, at least there's no examples. Now I remember using other async crate that let me manipulate TAR archives with different compression format, but I don't have access to that code anymore (commercial thing).

I think it was [`async-compression`](https://crates.io/crates/async-compression).

nathansgithub: Hm, I tried enabling MIME types for those other formats with `zip`, and I couldn't get it to work for extracting those files. I did just add the `tar` crate so it supports uncompressed tar files now.

jackpot51 (approved):

PR #359, opened 2024-08-12
feat: add option to permanently delete files and folders

dengelt:
This PR implements deleting files and folders permanently, skipping the trash.
Works with either Shift+Delete as a shortcut, or with a context menu entry.

The context menu entry is gated behind a setting, since this is what Nautilus does. This detail is in the last commit, which can easily be removed if this is not wanted.

Before permanently deleting stuff, the user is presented with a "are you sure" type dialog.

resolves #259 
resolves #348

PR #372, opened 2024-08-12
style: promote idiomatic functional Rust

michalfita:
I know this is unsolicited request, but as you're proud in writing Cosmic Desktop in Rust would be nice to promote more idiomatic functional style with less mutability. I don't mind if you throw this away as unsuitable for your needs.

Additionally I expanded the menu row to increase space between menu item name and the shortcut as translated items were too tight visually.

Working on this, however, led me to form following questions:
* This code demonstrates a lot of tedious chiseling to built context menu, but doesn't see to use [`iced_aw`](https://github.com/iced-rs/iced_aw?tab=readme-ov-file). Is there reason to build UI this way? Is there a roadmap what `libcosmic` would include in terms of standard desktop widgets?
* I found it surprising to keep mapping between _key binding_ and _action_ kept in `HashMap<>` to be iterated over to find _key binding_ for given _action_. Would it make sense to crate a type like `UIActionMap` that would hold some faster hash maps for both directions? For several key bindings in the application actually a vector with iterative search is faster than HashMap as actual hashing and bucketing takes plenty of cycles.
* Is the font final choice?

### DIFF
--- a/src/menu.rs
+++ b/src/menu.rs
@@ -75,49 +75,7 @@ pub fn context_menu<'a>(
         .into()
     };
 
-    struct SelectionCounter {
-        total_count: usize,
-        dirs_count: usize,
-    }
-
-    impl SelectionCounter {
-        fn new() -> Self {
-            Self {
-                total_count: 0,
-                dirs_count: 0,
-            }
-        }
-
-        fn any(&self) -> bool {
-            self.total_count > 0
-        }
-
-        fn exactly_one(&self) -> bool {
-            self.total_count == 1
-        }
-
-        fn exactly_one_dir(&self) -> bool {
-            self.dirs_count == 1
-        }
-
-        fn no_dirs(&self) -> bool {
-            self.dirs_count == 0
-        }
-
-        fn only_directories(&self) -> bool {
-            self.total_count == self.dirs_count
-        }
-    }
-
-    let selected = tab.items_opt().map_or(SelectionCounter::new(), |items| {
-        items.into_iter()
-            .filter(|i| i.selected)
-            .fold(SelectionCounter::new(), |mut counter, selection| {
-                counter.total_count += 1;
-                selection.metadata.is_dir().then(|| counter.dirs_count += 1);
-                counter
-            })
-    });
+    let selected = tab.selection_counts();
 
     let mut children: Vec<Element<_>> = Vec::with_capacity(16);
     match tab.location {
--- a/src/menu.rs
+++ b/src/menu.rs
@@ -109,14 +109,15 @@ pub fn context_menu<'a>(
         }
     }
 
-    let selected = tab.items_opt().into_iter().fold(SelectionCounter::new(),
-        |mut selections, items| {
-            let selected_iter = items.into_iter().filter(|i| i.selected);
-            selections.total_count += selected_iter.clone().count();
-            selections.dirs_count += selected_iter.filter(|i| i.metadata.is_dir()).count();
-            selections
-        }
-    );
+    let selected = tab.items_opt().map_or(SelectionCounter::new(), |items| {
+        items.into_iter()
+            .filter(|i| i.selected)
+            .fold(SelectionCounter::new(), |mut counter, selection| {
+                counter.total_count += 1;
+                selection.metadata.is_dir().then(|| counter.dirs_count += 1);
+                counter
+            })
+    });
 
     let mut children: Vec<Element<_>> = Vec::with_capacity(16);
     match tab.location {
--- a/src/menu.rs
+++ b/src/menu.rs
@@ -25,6 +25,7 @@ macro_rules! menu_button {
                 vec![$(Element::from($x)),+]
             )
             .align_items(Alignment::Center)
+            .spacing(8)
         )
         .height(Length::Fixed(32.0))
         .padding([4, 16])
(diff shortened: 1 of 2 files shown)


leb-kuchen:
Congrats, you made the code less readable and extensible.

michalfita:
@leb-kuchen I'd be glad if you could elaborate on that.

src/menu.rs
```diff
@@ -77,20 +75,16 @@ pub fn context_menu<'a>(
...
-                if item.metadata.is_dir() {
-                    selected_dir += 1;
-                }
-            }
+    let (selected, selected_dir) = tab.items_opt().into_iter().fold((0, 0),
+        |selections, items| {
+            let selected_iter = items.into_iter().filter(|i| i.selected);
+            let selected_count = selected_iter.clone().count();
+            let selected_dirs_count = selected_iter.filter(|i| i.metadata.is_dir()).count();
+            (selections.0 + selected_count, selections.1 + selected_dirs_count)
```
Kartonrealista, line 83: This replaces one for loop with two

Kartonrealista: Actually three loops, `selected_iter`, `selected_count` and `selected dirs_count`

michalfita: It'd been proven many times Rust's functional style gives compiler much wider field for optimisations than traditional imperative loops, especially if mutability is removed.

I can't run this through [godbolt](https://godbolt.org/) as they don't have `criterion` available, but my benchmark for 1000000 items, shows **2.7281 ms** for `for` loop and **471.58 µs** for `filter()/count()` iterators on my notebook PC. For 10000 difference is smaller, but still `filter()/count()` wins: 6.3825 µs / 4.7141 µs.

Style wise, I'd personally replace `tuple` with own type supporting addition to right fields and logic queries, etc.

leb-kuchen: This now O(n^2) instead of O(n).

leb-kuchen: Wait, are you folding over an option

leb-kuchen:
Go write Haskell

joshuamegnauth54:
I've had similar questions when I first started contributing to COSMIC. Don't take these answers as official. They're simply how I understand the situation.

>This code demonstrates a lot of tedious chiseling to built context menu, but doesn't see to use iced_aw.

[iced_aw](https://github.com/iced-rs/iced_aw/blob/18cf81403b19924164218581dfb8042f3de8096c/Cargo.toml#L104-L109) and [libcosmic](https://github.com/pop-os/libcosmic/blob/c9f8f485373e02ef849c83d3bbdf0fdd2bf4a24c/Cargo.toml#L180) depend on different versions of `iced` so (I think) it'd be difficult to use for COSMIC until the two branches of `iced` converge.

> Is there reason to build UI this way? Is there a roadmap what libcosmic would include in terms of standard desktop widgets?
 `libcosmic` implements its own set of widgets which are preferred because the team has full control over them. `iced` is a toolkit to build toolkits, so it is comparatively Spartan in relation to GUI libraries like GTK or Qt. I'm not sure if there's a roadmap for widgets that the library will provide, but there are open issues for widget improvements or new widgets in the `libcosmic` repo itself.

>but as you're proud in writing Cosmic Desktop in Rust would be nice to promote more idiomatic functional style with less mutability
[...]
>I found it surprising to keep mapping between key binding and action kept in HashMap<> to be iterated over to find key binding for given action. Would it make sense to crate a type like UIActionMap that would hold some faster hash maps for both directions? For several key bindings in the application actually a vector with iterative search is faster than HashMap as actual hashing and bucketing takes plenty of cycles.

COSMIC's development has been extremely fast. Some of the code quality issues you've mentioned is due to that alacritous dev time where finished features or bug fixes took precedence over tightly bound code. You'll notice throughout most of the codebases that there are TODOs to move certain implementations over to `libcosmic` or to update the code to use new `libcosmic` features. There are instances of duplicated features (e.g. localization or the key bindings) across the codebases that will likely be fixed later as the team figures out permanent solutions. I'm a noob so I tend to leave questions like that to the main developers. :joy:

src/menu.rs
michalfita, line 1: | Method | Avg t @ 10000 elements |
|------------|----------------------|
| original `for` | 6.4166 µs |
| `filter()`/`count()` (supposed O(n²) |  4.5204 µs |
| `filter()`/`fold()` (last one) |  3.6868 µs |

src/menu.rs
```diff
@@ -77,80 +75,76 @@ pub fn context_menu<'a>(
...
+                if selected.only_directories() {
                     children.push(menu_item(fl!("open-in-new-tab"), Action::OpenInNewTab).into());
                     children
                         .push(menu_item(fl!("open-in-new-window"), Action::OpenInNewWindow).into());
                 }
-                children.push(container(horizontal_rule(1)).padding([0, 8]).into());
-                children.push(menu_item(fl!("rename"), Action::Rename).into());
-                children.push(menu_item(fl!("cut"), Action::Cut).into());
-                children.push(menu_item(fl!("copy"), Action::Copy).into());
+                children.extend(vec![
```
dengelt, line 100: The `vec!` macro is not needed here, plain arrays implement `IntoIterator` as well. 

Also, wouldn't the use of `extend` versus a series of pushes make it less convenient to conditionally add menu items, like (currently pending) PRs like #359 and #370 do?

I guess I don't see the benefit of using `extend`. If it is to reduce the number of re-allocations as the `children` vec grows, this should already be solved by initializing with `with_capacity`.

dengelt: Personally, for what the code does now this would be too much abstraction. After all, right now all the builder would do is add elements to a vector. I don't know what the "minimum number of builder methods to warrant a builder pattern" would be, but I think it's more than 2. 

Also if I think about other places where builders are used, it's usually places where the creation is a complex process that needs to preserve certain invariants for every possible construction, and where different users create differently built objects. But here, there would only be one call to `build()`, because this is a very internal use, compared to builders that are part of the public API of HTTP libraries, database query builders, command line parsers, etc...

If this was in libcosmic, this should definitely be a builder pattern.

But ultimately, this is a decision someone from the pop-os team should make.

michalfita:
I got the feedback for my proposal that I considered and I have some ideas how to tackle certain things in this code to make it more long term maintainable, but I have to start from scratch.

PR #408, opened 2024-08-19
Rename sv.ftl to cosmic_files.ftl

bittin:
Rename the file so the Swedish translation works
as talked with about f0x on the Mattermost 

sorry i screwed up :(


bittin:
and maybe the folder should be named sv instead of sv-SE to be picked up, sorry for the hassle

jackpot51 (approved):

bittin:
@jackpot51 not sure if you have to rename the name of the folder from sv_SE to sv too 

maybe it would be good to rename it to sv and keep that as a standard for the Swedish translations
sorry for using the wrong folder name if that is a problem?

bittin:
@jackpot51 more Swedish then no Swedish now atleast but still 9-11 strings that are missing so guess i have to fix that after eating lunch 😛

bittin:
@jackpot51 see PR: #434

Issue #325, opened 2024-08-08
Complete the "Show details..." (Properties) context drawer


WatchMkr:
Complete per design with "Open with" and "Permissions" interfaces and "Open" button
When type=image add pixel width and height

![image](https://github.com/user-attachments/assets/0e5f3fc2-b43b-4463-815d-2ac02e16ff62)

[Design File](https://www.figma.com/design/ZBg577prx1KF8BlZrBSgT0/Files?node-id=2847-25477&t=FhDYB7EXyARkip9u-0)

WatchMkr:
Should include the option to make executable.

WatchMkr:
That all sounds good.

Omnimaxus:
Please add the option to make files executable.  I don't want to have to use chmod each and every time in the terminal when I want to install a GOG game using their .sh downloads.  Thank you very much, System76.

PR #414, opened 2024-08-21
Display file permissions in metadata list

Fingel:
Displays the three digit permissions of a file in the sidebar under the other metadata. 

This seemed like an omission, the other metadata listed here made me expect to also find file permissions. I've truncated the bits down to the last 3, which is the set I think most people are interested in.

Digit based display:
image

Character based display:
image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1261,6 +1261,7 @@ dependencies = [
  "test-log",
  "tokio",
  "trash",
+ "unix_permissions_ext",
  "url",
  "vergen",
  "xdg",
@@ -1530,7 +1531,7 @@ checksum = "67e77553c4162a157adbf834ebae5b415acbecbeafc7a74b0e886657506a7611"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.72",
+ "syn 2.0.75",
 ]
 
 [[package]]
@@ -6094,6 +6095,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f962df74c8c05a667b5ee8bcf162993134c104e96440b663c8daa176dc772d8c"
 
 [[package]]
+name = "unix_permissions_ext"
+version = "0.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7497808a85e03f612f13e9c5061e4c81cdee86e6c00adfa1096690990ccd08e9"
+
+[[package]]
 name = "url"
 version = "2.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


jacobgkau:
Long-term, it might be good to try and display this in a more human-readable format for people who don't know what the numbers mean? Or at least have a parenthetical with it in e.g. rw-r--r-- format? The other devs and/or UX team may have ideas.

I think this would be nice to have as an option, though.

edfloreshz:
Just a suggestion here, given that a lot of non technical people might end up using cosmic-files, we might want to display permissions in a more user friendly manner.

644 tell us, the Linux experienced users, what we need to know, but not everybody has that knowledge.

jacobgkau:
Since I'm in front of KDE right now, I'll share what this looks like in Dolphin. It's not shown by default, but Owner, Group, and Permissions can be configured to display with the letters format:
Screenshot_20240820_223519
Still not as straightforward as it could be. (I'd probably list "read", "read/write", etc underneath the owner, the group, and have one more section for "other.")

Fingel:
There are a few [crates](https://docs.rs/crate/unix_permissions_ext/latest) that would make changing the display trivial. It seems a bit silly to pull in a third party dependency for such a small thing, but I fear I couldn't do a better job myself.

Fingel:
I added a commit to use the `unix_permissions_ext` crate to provide a letter based display.
image

I agree that the most friendly display would probably be something like:
user: austin (read/write)
group: austin (read)
other: (read)

Though this is a bit of scope increase as it also adds display for the user and group which would also be a nice addition. I can look into that tomorrow.

edfloreshz:
This is great! It could even be added as an option to display either the code or the formatted string. 

Great job! 👏

leb-kuchen:
I do not think a dependency should be used for something, which is as simple as this. 
https://cs.opensource.google/go/go/+/refs/tags/go1.23.0:src/io/fs/fs.go;l=201
unix_permissions_ext, apart from  the nine least-significant, only supports, setuid, setgid and the sticky bit.

leb-kuchen:
Here is an example implementation without dependencies and with no heap allocations.
This does not list file types like ls, nor does unix_permissions_ext, even though the crate says it is identical to ls(1).
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=847a2b947d9d463d39da60d13c779047

jackpot51:
We have designs for this here: https://github.com/pop-os/cosmic-files/issues/325

Fingel:
> We have designs for this here: #325

Ah, RIP. Well I supposed this applies to #421 as well. That one is closer but it looks like ya'll have a plan.

leb-kuchen:
The file permissions bits could still be show, if the user is not the file owner.

PR #421, opened 2024-08-21
Display file ownership and permissions.

Fingel:
This subsumes #414. Took feedback into account to display more human-readable permissions. Also, display the owner and group ownership inline with the permissions themselves. 

Decided to put this in a new PR in case the old one is preferred, this one has a bit more going on.

image

### DIFF
--- a/src/tab.rs
+++ b/src/tab.rs
@@ -43,6 +43,7 @@ use std::{
     fmt,
     fs::{self, Metadata},
     num::NonZeroU16,
+    os::unix::fs::MetadataExt,
     path::PathBuf,
     sync::{Arc, Mutex},
     time::{Duration, Instant},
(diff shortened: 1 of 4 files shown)


leb-kuchen:
Instead of using a dependency and pattern matching you can use this function. You can make PermsissionOwner a repr 8. For operation, 1 means read, 2 write and 3 exec.

Essentially this treats mode as an array of bool, and calculates the right index. 

```rs
fn permission_owner_has_permission(mode: u32, owner: u8, operation: u8) -> bool {
  mode & ( 1 <<  (9 - 3 * owner - operation)) != 0
}
```

src/tab.rs
```diff
@@ -199,6 +202,60 @@ fn format_size(size: u64) -> String {
...
+        PermissionOwner::Group => metadata.permissions().writable_by_group(),
+        PermissionOwner::Other => metadata.permissions().writable_by_other(),
+    };
+    let executable = match owner {
+        PermissionOwner::Owner => metadata.permissions().executable_by_owner(),
+        PermissionOwner::Group => metadata.permissions().executable_by_group(),
+        PermissionOwner::Other => metadata.permissions().executable_by_other(),
+    };
+    format!(
+        "{} {} {}",
```
jackpot51, line 239: I believe if this is readable and executable but not writable, there will be an extra space between Read and Execute.

jackpot51: One way to fix this would be to push the read, write, and execute strings to a Vec and use join(' ') to turn it into a string.

Fingel: Thanks for the review! Bug fixed, and the code is _much_ cleaner now.

jackpot51 (approved):

PR #428, opened 2024-08-29
hotkeys-capslock fix for (#286)

### LINKED ISSUE #286: Hotkeys: capslock being active breaks a bunch of hotkeys
If I turn on capslock, I notice that hotkeys no longer work. The ones I've noticed so far:
- Ctrl + Shift + N to create new folder
- Ctrl + T to open a new tab
- Ctrl + N to open a new window
- Ctrl + W to close a tab or window

Possibly more, but these are the quickest ones I could recreate

l-const:
Should close #286 .
Demo:

[hotkeys-capslock.webm](https://github.com/user-attachments/assets/89bc358a-0534-4790-87d7-570ba734c278)

Note: I did try to change the macro to add nested repetition but the metavariables need to have same length to work according to the reference: https://doc.rust-lang.org/reference/macros-by-example.html#transcribing.

> If multiple metavariables appear in the same repetition, they must be bound to the same number of fragments.

### DIFF
--- a/src/key_bind.rs
+++ b/src/key_bind.rs
@@ -24,9 +24,13 @@ pub fn key_binds() -> HashMap<KeyBind, Action> {
     }
 
     bind!([Ctrl], Key::Character("d".into()), AddToSidebar);
+    bind!([Ctrl], Key::Character("D".into()), AddToSidebar);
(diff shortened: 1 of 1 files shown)


jackpot51:
I don't think this is the best way to resolve this. We should be able to match character keys without relying on their case.

PR #434, opened 2024-09-03
Update cosmic_files.ftl Swedish translation

bittin:
Updated Swedish Translation 
(was unsure if the product name COSMIC Files should be Cosmic Filer instead in Swedish or not or if you translate product names or not?)

### DIFF
--- a/i18n/sv/cosmic_files.ftl
+++ b/i18n/sv/cosmic_files.ftl
@@ -60,7 +60,7 @@ select-all = Välj alla
 
 ## View
 zoom-in = Zooma in
-default-size = Standard storlek
+default-size = Standardstorlek
 zoom-out = Zooma ut
 view = Visa
 grid-view = Rutnätsvy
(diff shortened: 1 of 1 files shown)


i18n/sv/cosmic_files.ftl
```diff
@@ -57,7 +59,20 @@ paste = Klistra in
 select-all = Välj alla
 
 ## View
+zoom-in = Zooma in
+default-size = Standard storlek
+zoom-out = Zooma ut
```
Tebro, line 64: Should the zoom ones be "Förstora" and "Förminska" instead of Zoom based?

haaninjo: No hard opinion there, but "Zooma in/ut" is a quite common translation in other places.

I would write "Standardstorlek" as one word though.

bittin: Fixed that now, sorry for särskrivningen

jackpot51 (approved):

PR #436, opened 2024-09-03
Add Compress menu option to add files to zip archive

nathansgithub:
Continuing the work I did on https://github.com/pop-os/cosmic-files/commit/ef23ff77e1442cb348901f8d6720341fb0da0685 (an Extract archive option) for https://github.com/pop-os/cosmic-files/issues/122. This PR adds a right-click option to compress files into a zip archive in the current directory, and lets you choose a new file name.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -67,6 +67,7 @@ pub struct Flags {
 pub enum Action {
     About,
     AddToSidebar,
+    Compress,
     Copy,
     Cut,
     EditHistory,
@@ -116,6 +117,7 @@ impl Action {
         match self {
             Action::About => Message::ToggleContextPage(ContextPage::About),
             Action::AddToSidebar => Message::AddToSidebar(entity_opt),
+            Action::Compress => Message::Compress(entity_opt),
             Action::Copy => Message::Copy(entity_opt),
             Action::Cut => Message::Cut(entity_opt),
             Action::EditHistory => Message::ToggleContextPage(ContextPage::EditHistory),
@@ -210,6 +212,7 @@ pub enum Message {
     AddToSidebar(Option<Entity>),
     AppTheme(AppTheme),
     CloseToast(widget::ToastId),
+    Compress(Option<Entity>),
     Config(Config),
     Copy(Option<Entity>),
     Cut(Option<Entity>),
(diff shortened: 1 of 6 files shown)


src/app.rs
```diff
@@ -2312,6 +2339,66 @@ impl Application for App {
...
+                    dialog = dialog.tertiary_action(widget::text::body(fl!(
+                        "name-invalid",
+                        filename = name.as_str()
+                    )));
+                    None
+                } else if name.contains('/') {
+                    dialog = dialog.tertiary_action(widget::text::body(fl!("name-no-slashes")));
+                    None
+                } else {
+                    let path = to.join(name);
```
jackpot51, line 2357: This does not test the existence properly, as the name does not have .zip added

src/app.rs
```diff
@@ -1264,6 +1272,20 @@ impl Application for App {
...
             }
+            Message::Compress(entity_opt) => {
+                let paths = self.selected_paths(entity_opt);
+                if let Some(current_path) = paths.get(0) {
+                    if let Some(destination) = current_path.parent().zip(current_path.file_stem()) {
+                        let to = destination.0.to_path_buf();
+                        self.dialog_pages.push_back(DialogPage::Compress {
+                            paths,
+                            to,
+                            name: String::new(),
```
jackpot51, line 1283: It would be nice to pre-fill the name. If there is only one path selected, it should probably be that path with its extension replaced with .zip

src/app.rs
nathansgithub: Good catch. I just added the extension here.

src/operation.rs
```diff
@@ -348,6 +365,68 @@ impl Operation {
...
+                                if path.is_dir() {
+                                    let new_paths_it = WalkDir::new(path).into_iter();
+                                    for entry in new_paths_it.skip(1) {
+                                        let entry = entry.map_err(err_str)?;
+                                        paths.push(entry.path().to_path_buf());
+                                    }
+                                }
+                            }
+
+                            for path in paths {
```
jackpot51, line 394: Can progress be reported? Currently it just goes from 0% straight to 100% when it finishes.

jackpot51:
This looks really good now! I just have one more comment.

nathansgithub:
> This looks really good now! I just have one more comment.

Thank you! I just added the per-file progress. I also went ahead and did the same thing for the Extract action, and have it showing the completed message for both of those.

jackpot51 (approved):

Issue #252, opened 2024-06-17
Deletion of files in "find" view doesn't update view


flukejones:
Per title really.

1. Do a `ctrl+f` for search, and search for something to delete
2. Select any of these to delete and do so
3. The view does not update to reflect removal
4. Select the files and try to delete again to see error message
5. Do a search again for the same files and see that they really were deleted

l-const:
Should be closed by: https://github.com/pop-os/cosmic-files/pull/437

PR #437, opened 2024-09-03
Delete files in search-mode. (fix for #252)

l-const:
Should close https://github.com/pop-os/cosmic-files/issues/252  (#252 )

Demo:

[delete-in-search-mode.webm](https://github.com/user-attachments/assets/5a11f245-14cf-475e-82b3-0e7d58466384)

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -1696,6 +1696,11 @@ impl Application for App {
                 commands.push(self.update_notification());
                 // Manually rescan any trash tabs after any operation is completed
                 commands.push(self.rescan_trash());
+
+                // if search is active, update "search" tab view
+                if !self.search_input.is_empty() {
+                    commands.push(self.search());
+                }
                 return Command::batch(commands);
             }
             Message::PendingError(id, err) => {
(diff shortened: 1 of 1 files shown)


src/app.rs
```diff
@@ -1842,8 +1847,14 @@ impl Application for App {
                 if let Some(tab) = self.tab_model.data::<Tab>(entity) {
                     self.activate_nav_model_location(&tab.location.clone());
                 }
-
-                return self.update_title();
+                let mut commands = vec![];
+                commands.push(self.update_title());
+                // if the tab was in an active search mode
```
jackpot51, line 1852: TabActivate happens when a tab is clicked on. I'm not sure why this needs to rerun the search.

l-const: There is the issue when multiple search tabs are open (with the same search input or slightly different that could have a joint set of search result), so what i did since we do not have interactive search as gnome-files is to research when you activate a tab that had a search active.

jackpot51: This is good enough for now, I suppose.

PR #441, opened 2024-09-04
Refresh/reload tab view & search, fix(#146).

### LINKED ISSUE #146: No option to refresh current folder (F5 key and/or menu)
F5 isn't refreshing the folder.

l-const:
- Should close #146 
 
- Made an attempt to provide a refresh/reload view functionality and it seems to work but maybe not the best/novel way to do it. I managed to create a refresh-delay animation somehow otherwise the reload functionality would not be able to be noticed, similar to how nautilus at least from a user perspective is doing that.
 
- Demo: 

 [refresh-view.webm](https://github.com/user-attachments/assets/8cac4b9c-5328-4fbb-b11c-ffe15505845c)

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -69,6 +69,8 @@ use crate::{
     tab::{self, HeadingOptions, ItemMetadata, Location, Tab, HOVER_DURATION},
 };
 
+static REFRESH_VIEW_DELAY_MS: u64 = 20; 
+
 #[derive(Clone, Debug)]
 pub enum Mode {
     App,
@@ -111,6 +113,7 @@ pub enum Action {
     OpenWith,
     Paste,
     Properties,
+    Refresh,
     Rename,
     RestoreFromTrash,
     SearchActivate,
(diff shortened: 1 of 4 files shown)


src/app.rs
```diff
@@ -57,6 +57,8 @@ use crate::{
     tab::{self, HeadingOptions, ItemMetadata, Location, Tab, HOVER_DURATION},
 };
 
+static REFRESH_VIEW_DELAY_MS: u64 = 20;
```
l-const, line 60: The delay value was a totally arbitrary decision.

src/app.rs
```diff
@@ -1717,6 +1724,28 @@ impl Application for App {
                 }
                 return self.update_notification();
             }
+            Message::Refresh(entity_opt) => {
+                let entity = entity_opt.unwrap_or_else(|| self.tab_model.active());
+                if let Some(tab) = self.tab_model.data_mut::<Tab>(entity) {
+                    let location = tab.location.clone();
+                    tab.set_items(vec![]);
```
l-const, line 1731: Essentially, I set an "empty" value for the given tab and register a delayed callback that will either rescan or search again. Maybe something better could be done here.

l-const:
I will try to add a new title and perhaps icon for the in between "Loading..." view or something like that. I am not sure where i could find a "Loading..." icon.

l-const:
Attaching a new demo with the "Loading..." view, note that for demonstrative purposes the delay in the video was set at 200ms and not 20ms the one that is commited. 

Demo: 
[refresh-loading-200ms.webm](https://github.com/user-attachments/assets/ec72bd06-3e9c-4de1-ab3a-d8789a1bd825)

src/app.rs
l-const: I guess in a 60Hz display you need at least 16.6 ms , so you will see at least one frame  and more that one in higher refresh rate displays assuming the execution of the rescan of the view contents is instantaneous.

iamkartiknayak:
Has there been any dev regarding this? This is a necessary feature

PR #444, opened 2024-09-08
Vertically center context menu items

git-f0x:
Makes the context menu items vertically centered.
Also removes the hacky list view item padding (that I put there) and makes the line under the breadcrumb accent-colored.
Edit: Not sure if the `.padding([0, space_xxs])` in list view should possibly use 8 instead of `space_xxs`, since that padding probably shouldn't change with interface density settings.


git-f0x:
The spacing variables should probably be somehow included in the context menu. Should maybe the item height be something like `20.0 + space_xs` if possible (padding would have the variables doubled, so might not be the best option for more fine-grained control), so that it changes with interface density settings?

jackpot51:
This needs a rebase since the list of supported archives changed. It looks good so I will merge after that.

git-f0x:
Sorry for the multiple commits. I don't have much experience with merge conflicts.

PR #451, opened 2024-09-11
open multiple files in one app instance if possible

elevenhsoft:
This will allow to open multiple files in one app instance. So for example, we opening 3x text files, so they will be opened in COSMIC Text (default app for specific mimetype) in 3 tabs, instead of 3 apps. Don't know if this is needed or implementation is fine but I'm sure to not broke anything :sweat_smile: If there will be one file selected, it will open just like before. Only handle for multiple files is featured. I tested this on multiple files for the same type:
- 3x text files: *.rs, *.md and *.txt = opened in 3 tabs of COSMIC Text
- 2x files with the same mimetype, *.deb = opened in one instance of [Wizard](https://github.com/cosmic-utils/wizard) app, so expected (also tested it in this app ^^)
- 4x files with 2 different mimetypes, 2x *.deb and 2x *.zip files = opened in 2x different fileroller instances (so as it should I guess) and 2x deb files opened in one instance of Wizard app.

From my testing, seems to work.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -1677,7 +1677,7 @@ impl Application for App {
                         }
                     }
                     for path in paths {
-                        if let Some(mut command) = terminal.command(None) {
+                        if let Some(mut command) = terminal.command(Vec::new()) {
                             command.current_dir(&path);
                             match spawn_detached(&mut command) {
                                 Ok(()) => {}
@@ -1697,7 +1697,7 @@ impl Application for App {
                 }
             }
             Message::OpenWith(path, app) => {
-                if let Some(mut command) = app.command(Some(path.clone())) {
+                if let Some(mut command) = app.command(vec![Some(path.clone())]) {
                     match spawn_detached(&mut command) {
                         Ok(()) => {
                             let _ = recently_used_xbel::update_recently_used(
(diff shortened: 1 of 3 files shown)


src/mime_app.rs
```diff
@@ -10,16 +10,20 @@ use std::{
...
 
-pub fn exec_to_command(exec: &str, path_opt: Option<PathBuf>) -> Option<process::Command> {
+pub fn exec_to_command(exec: &str, path_opt: Vec<Option<PathBuf>>) -> Option<process::Command> {
     let args_vec: Vec<String> = shlex::split(exec)?;
     let mut args = args_vec.iter();
     let mut command = process::Command::new(args.next()?);
     for arg in args {
         if arg.starts_with('%') {
             match arg.as_str() {
                 "%f" | "%F" | "%u" | "%U" => {
```
jackpot51, line 20: If a program provides %f or %u it means only a single path or URL. %F or %U can accept a list.

src/app.rs
```diff
@@ -2165,6 +2165,51 @@ impl Application for App {
                                 }
                             }
                         }
+                        tab::Command::OpenMultipleFiles(items) => {
```
jackpot51, line 2168: There should be a fallback to the original open code if no default apps are found, which will always happen on OSes like macOS and Windows.

src/mime_app.rs
elevenhsoft: Oh, did not know. To be honest I was reading about it today, because testing something but I did not consider it here :sweat: Should I split this match pattern and in this case provide only first argument from vector and full list in other case?

src/app.rs
elevenhsoft: Oops, did not think about it. Of course I will make this change :)

elevenhsoft:
I made some changes.
First, now checking if Item has default app to open with. If it has, it will push command to open multiple files, otherwise it will use OpenFile command. I think it should works with different OSes but not sure. Someone smarter have to take a look ^^

Second, about arguments. Was not sure how to handle this case, so just passing first argument from vector. I don't know if it's correct tho.

Issue #343, opened 2024-08-09
Easier to read modified date and time


WatchMkr:

maria-komarova:
![Home - list](https://github.com/user-attachments/assets/8096452f-8493-49c3-ac3c-49566da69f76)
- Use "today" when the item has been modified starting 12am of this date.
- Remove seconds
- Use Month followed by date, separated with comma from the year.

PR #464, opened 2024-09-15
Change display of date and time

jasonrhansen:
Implements  #343

It currently assumes a 12-hour clock for the time format. It would probably make sense to add a setting to be able to switch it to a 24-hour clock. Maybe it could just use whatever is configured for the time applet?


jackpot51:
I agree on using the same setting the clock widget uses.

jackpot51 (approved):

jackpot51:
The 24 hour format can be done in a follow up, thanks for this!

PR #520, opened 2024-09-28
Fix extraction folder naming issues and handle existing folders correctly

### LINKED ISSUE #504: [bug] Extracting an archive with multiple . characters in the filename
**Cosmic-files version:** 1.0.0.alpha.1.r170.g8668c18-1

**Issue/Bug description:** Extracting a file with multiple `.` characters in the filename (e.g. Archipelago_0.5.0_linux-x86_64.tar.gz) creates a folder with a cut-off name (Archipelago_0)

**Steps to reproduce:** Right-click an archive with a name such as the example, extract it

**Expected behavior:** The created folder should match the entire name minus the archive's extension (in this case Archipelago_0.5.0_linux-x86_64)

iamkartiknayak:
### Overview
This pull request addresses two related issues in the extraction functionality of the application:

1. **Preserving Full Folder Names**: 
   Previously, extracted folders with names containing periods before the extension (e.g., `file.last.6.4.tar.gz`) were being incorrectly named, leading to truncated folder names (e.g., `file`). This fix ensures that the full original name is preserved during extraction.

2. **Correctly Appending Copy Suffix**:
   A bug was identified where, if a folder with the same name already existed, the new folder was incorrectly named with the copy suffix placed in the wrong location. For instance, a folder named `Test_File_0.5.0last` would create a new folder as `Test_File_0.5 (Copy 1).0last` instead of the expected `Test_File_0.5.0last (Copy 1)`. This issue has been resolved to append the copy suffix correctly.

### Related Issues
- Fixes #504

### Testing
- Please verify that extraction preserves the correct folder names for files with complex extensions and that copy suffixes are appended correctly for existing folders.

### DIFF
--- a/src/operation.rs
+++ b/src/operation.rs
@@ -330,30 +330,21 @@ async fn copy_or_move(
 
 fn copy_unique_path(from: &Path, to: &Path) -> PathBuf {
     let mut to = to.to_owned();
-    // Separate the full file name into its file name plus extension.
-    // `[Path::file_stem]` returns the full name for dotfiles (e.g.
-    // .someconf is the file name)
-    if let (Some(stem), ext) = (
-        // FIXME: Replace `[Path::file_stem]` with `[Path::file_prefix]` when stablized to handle .tar.gz et al. better
-        from.file_stem().and_then(|name| name.to_str()),
-        from.extension()
-            .and_then(|ext| ext.to_str())
-            .unwrap_or_default(),
-    ) {
-        // '.' needs to be re-added for paths with extensions.
-        let dot = if ext.is_empty() { "" } else { "." };
+    if let Some(file_name) = from.file_name().and_then(|name| name.to_str()) {
         let mut n = 0u32;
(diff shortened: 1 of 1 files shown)


jackpot51:
I have comcerns the changes to making a unique name break extensions on copied files. Copying `test.txt` should create `test (Copy 1).txt`

iamkartiknayak:
Thanks for the review. I have fixed it.

iamkartiknayak:
Implemented a fix but it only works for copying
- dir 
- file.ext

Doesn't work for files with more than 1 ext such as `tar.gz` or `tar.bz2`

https://github.com/user-attachments/assets/29086cac-4476-4919-8008-41370c661e14

iamkartiknayak:
> Doesn't work for files with more than 1 ext such as `tar.gz` or `tar.bz2`

I'll work on this later.

jackpot51 (approved):

PR #521, opened 2024-09-28
Run executables

### LINKED ISSUE #23: [Feature Request] Run executables
Many file managers allow launching executables from them.
For example I have a few appimages in a location where launching them from a GUI application is faster than a terminal

francesco-gaglione:
Closes #23 #525

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -3136,7 +3136,7 @@ impl Application for App {
                             .on_press(Message::DialogComplete),
                     )
                     .secondary_action(
-                        widget::button::text("cancell")
+                        widget::button::text(fl!("cancel"))
                             .style(theme::Button::Destructive)
                             .on_press(Message::DialogCancel),
                     )
--- a/src/app.rs
+++ b/src/app.rs
@@ -256,6 +256,7 @@ impl MenuAction for NavMenuAction {
 pub enum Message {
     AddToSidebar(Option<Entity>),
     AppTheme(AppTheme),
+    AddExecutablePermission(PathBuf),
     CloseToast(widget::ToastId),
     Compress(Option<Entity>),
     Config(Config),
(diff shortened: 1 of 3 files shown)


src/app.rs
```diff
@@ -2350,18 +2342,51 @@ impl Application for App {
...
-                                        log::warn!("failed to open {:?}: {}", path, err);
+                                let file_extension = path.extension();
+                                match file_extension {
+                                    Some(ext) if ext == OsStr::new("AppImage") => {
+                                        // Set the executable permission to the file
+                                        let mut perms = fs::metadata(&path)
+                                            .expect("Failed to get metadata")
+                                            .permissions();
+                                        // Set the executable permission to the file
+                                        perms.set_mode(0o755);
```
jackpot51, line 2353: This should involve a dialog to ask to mark it as executable. Then the bits for executable should be ORed on to the current mode, rather than indiscriminately setting the mode to 755.

src/app.rs
```diff
@@ -2350,18 +2342,51 @@ impl Application for App {
...
+                                        fs::set_permissions(&path, perms).expect("Failed to set permissions");
+
+                                        log::info!("running app: {:?}", ext);
+                                        let cmd = std::process::Command::new(path).spawn();
+                                        match cmd {
+                                            Ok(mut res) => {
+                                                if let Some(stderr) = res.stderr.take() {
```
jackpot51, line 2360: stderr is not piped by default so this can be dropped.

iamkartiknayak:
Hey, Can you also add support for binaries? At present it opens in default browser and ask to save the file.
`.AppImage` is working  🚀

jackpot51:
@francesco-gaglione if you want, I can take over to add the dialog for enabling the executable permission, and for also triggering when the file is an ELF file rather than just when it has an AppImage extension.

iamkartiknayak:
Would it better to rename `Add permission` to Execute instead? Currently it's asking the permission even if AppImage has execute permission setup. 

`Cancel` button needs fix, it's misspelled

francesco-gaglione:
I Washington trying to do not ask for permission if it already have it. So in tratti case the nome is perfect but execution should be ouside the dialog. I will fix these for now, but later I want try to check run permission, ask if needed and than execute it (any executable). I didn't jad much time, if no one will finish before me Ill try asap

PR #565, opened 2024-10-04
hide files in .hidden file

### LINKED ISSUE #549: [FEATURE REQUEST] Support for `.hidden` file to hide files/directories like Nautilus
**Cosmic-files version:**
```
cosmic-files:
  Installed: 0.1.0~1727906839~24.04~dabb374
```

**Issue/Bug description:**
Add support for .hidden files in Cosmic Files, similar to the functionality in Nautilus. This feature would allow users to hide specific files and directories by listing them within a .hidden file per directory.

**Steps to reproduce:**
N/A

**Other notes:**

https://github.com/user-attachments/assets/905f0032-6b20-49b6-bbe6-ca1784b7f91a

francesco-gaglione:
Close #549

### DIFF
--- a/src/tab.rs
+++ b/src/tab.rs
@@ -447,6 +447,7 @@ pub fn item_from_path<P: Into<PathBuf>>(path: P, sizes: IconSizes) -> Result<Ite
 
 pub fn scan_path(tab_path: &PathBuf, sizes: IconSizes) -> Vec<Item> {
     let mut items = Vec::new();
+    let mut hidden_file = Vec::new();
     match fs::read_dir(tab_path) {
         Ok(entries) => {
             for entry_res in entries {
@@ -472,6 +473,10 @@ pub fn scan_path(tab_path: &PathBuf, sizes: IconSizes) -> Vec<Item> {
                     }
                 };
 
+                if name == ".hidden" {
+                    hidden_file = parse_hidden_file(&path);
+                }
+
                 let metadata = match fs::metadata(&path) {
                     Ok(ok) => ok,
                     Err(err) => {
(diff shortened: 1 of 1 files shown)


jackpot51 (changes requested):
Why not set item.hidden to true when it shows up in the .hidden file, in the scan function?

src/tab.rs
```diff
@@ -491,6 +497,17 @@ pub fn scan_path(tab_path: &PathBuf, sizes: IconSizes) -> Vec<Item> {
         (false, true) => Ordering::Greater,
         _ => LANGUAGE_SORTER.compare(&a.display_name, &b.display_name),
     });
+    items.iter_mut().for_each(|mut item| {
+        if item.path_opt().is_some() {
```
jackpot51, line 501: I would recommend to use if let Some(path) = item.path_opt() here, then use hidden.eq(path) below.

src/tab.rs
```diff
@@ -1358,6 +1375,35 @@ fn folder_name<P: AsRef<Path>>(path: P) -> (String, bool) {
...
+    };
+
+    let reader = BufReader::new(file);
+    let mut paths: Vec<PathBuf> = Vec::new();
+
+    for line in reader.lines() {
+        if let Ok(line) = line {
+            if !line.is_empty() {
+                let file_path = parent_dir.join(&line.trim());
+                if file_path.exists() {
```
jackpot51, line 1397: I don't think this check is necessary

src/tab.rs
```diff
@@ -1358,6 +1375,35 @@ fn folder_name<P: AsRef<Path>>(path: P) -> (String, bool) {
...
+        Err(_) => return Vec::new(),
+    };
+
+    let reader = BufReader::new(file);
+    let mut paths: Vec<PathBuf> = Vec::new();
+
+    for line in reader.lines() {
+        if let Ok(line) = line {
+            if !line.is_empty() {
+                let file_path = parent_dir.join(&line.trim());
```
jackpot51, line 1396: Converting to a full path is probably not necessary. You can store line.trim() in the Vec, and make it a Vec<String>, then test the item's name instead of path.

jackpot51 (approved):

PR #587, opened 2024-10-10
Update location edit selection behavior

nathansgithub:
This updates the keyboard behavior around editing the location to act more like other programs like Nautilus and Firefox. 
When you hit ctrl+l, it immediately allows you to type in the location box. You can de-select the location box by hitting escape or clicking in the tab window.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -153,7 +153,7 @@ impl Action {
             Action::DesktopViewOptions => Message::DesktopViewOptions,
             Action::EditHistory => Message::ToggleContextPage(ContextPage::EditHistory),
             Action::EditLocation => {
-                Message::TabMessage(entity_opt, tab::Message::EditLocationToggle)
+                Message::TabMessage(entity_opt, tab::Message::EditLocationEnable)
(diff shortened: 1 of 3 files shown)


src/app.rs
```diff
@@ -1469,6 +1469,11 @@ impl Application for App {
                 return Command::none();
             }
 
+            if tab.edit_location.is_some() {
```
jackpot51, line 1472: You'd probably also want to make this change in dialog.rs

jackpot51 (commented):
The edit button should still be able to cancel editing the path when clicked, in my opinion

nathansgithub:
> The edit button should still be able to cancel editing the path when clicked, in my opinion

Do you mean the pencil/x icon next to the file path? That still cancels editing the path for me.

I also just changed it to cancel editing when you right click on the nav menu, too. I couldn't figure out a good way to do the same if you open a right click menu from the title bar. But I think this is an improvement on the current behavior.

jackpot51 (approved):

PR #588, opened 2024-10-12
feat: enable LTO

zamazan4ik:
Hi!

I started a discussion about enabling Link-Time Optimization (LTO) across all pop!_os projects to improve their general performance (more compiler optimizations can be done with LTO) and the binary size reduction (LTO usually leads to measurable binary size improvements) here - https://github.com/pop-os/pop/discussions/3386 . In https://github.com/pop-os/pop/discussions/3386#discussioncomment-10909426 was proposed creating PRs into repos with enabling LTO - it's such a PR!

As a reference, I used the `cosmic-comp` Release [profile](https://github.com/pop-os/cosmic-comp/blob/master/Cargo.toml#L116).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -81,6 +81,9 @@ wgpu = ["libcosmic/wgpu"]
 [profile.dev]
 opt-level = 1
 
+[profile.release]
+lto = "fat"
+
 [profile.release-with-debug]
 inherits = "release"
 debug = true


Cargo.toml
```diff
@@ -81,6 +81,9 @@ wgpu = ["libcosmic/wgpu"]
 [profile.dev]
 opt-level = 1
 
+[profile.release]
+lto = "fat"
```
JohnRTitor, line 85: This lto performs better.

zamazan4ik: Did you mean Thin LTO?

JohnRTitor: Yes, I thought that was obvious.

Issue #310, opened 2024-08-01
[Bug] Cosmic files takes a long time to load if you have a lot of deleted files in the trash.


amix3:
**The Bug:**
Cosmic files taking a long time to load. it takes about 4 seconds for it to load completely.

**How to reproduce:**
Move a bunch of files to the trash can, 
I am able to reproduce it when I move papirus-icon-theme files to the trash can.

![screenshot-2024-08-01-11-20-29](https://github.com/user-attachments/assets/eeafb679-c7d1-4d65-8b39-e03b53b66503)

After doing this the file manager takes a long time to load and gives these logs when i run it in the terminal:

`
[2024-08-01T11:11:30Z ERROR i18n_embed::fluent] Error while adding resource to bundle: Overriding { kind: Message, id: "undo" }.
[2024-08-01T11:11:30Z WARN  sctk_adwaita::buttons] Ignoring unknown button type:
[2024-08-01T11:11:30Z WARN  wgpu_hal::gles::adapter] Detected skylake derivative running on mesa i915. Clears to srgb textures will use manual shader clears.
[2024-08-01T11:11:31Z WARN  iced_futures::subscription::tracker] Error sending event to subscription: TrySendError { kind: Full }
[2024-08-01T11:11:31Z WARN  iced_futures::subscription::tracker] Error sending event to subscription: TrySendError { kind: Full }
[2024-08-01T11:11:31Z WARN  iced_futures::subscription::tracker] Error sending event to subscription: TrySendError { kind: Full }
[2024-08-01T11:11:31Z WARN  iced_futures::subscription::tracker] Error sending event to subscription: TrySendError { kind: Full }
[2024-08-01T11:11:31Z WARN  iced_futures::subscription::tracker] Error sending event to subscription: TrySendError { kind: Full }
[2024-08-01T11:11:31Z WARN  iced_futures::subscription::tracker] Error sending event to subscription: TrySendError { kind: Full }
[2024-08-01T11:11:31Z WARN  iced_futures::subscription::tracker] Error sending event to subscription: TrySendError { kind: Full }
[2024-08-01T11:11:31Z WARN  iced_futures::subscription::tracker] Error sending event to subscription: TrySendError { kind: Full }
[2024-08-01T11:11:31Z WARN  iced_futures::subscription::tracker] Error sending event to subscription: TrySendError { kind: Full }
[...]

PR #604, opened 2024-10-16
perf: Only enumerate trash items when necessary [Waiting on trash-rs patch]

joshuamegnauth54:
Closes: #310

Depends on Byron/trash-rs#120 which implements a function that only checks if the trash is empty or not instead of parsing each item.

### DIFF
--- a/src/tab.rs
+++ b/src/tab.rs
@@ -204,7 +204,7 @@ pub fn trash_entries() -> usize {
 }
 
 pub fn trash_icon(icon_size: u16) -> widget::icon::Handle {
-    widget::icon::from_name(if trash_entries() > 0 {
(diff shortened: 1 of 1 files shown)


dtomvan:
@joshuamegnauth54 The PR has been merged on `trash-rs`'s end, so this can go through now, yes? Could you do a cargo-update and open this for review?

joshuamegnauth54:
Thanks! I didn't notice that the patch has been merged.

I think that COSMIC Files uses a fork of `trash-rs`, so I'll have to see if I can rebase that on main first and then set this PR ready.

jackpot51:
@joshuamegnauth54 the fork is up to date, do you want to finish this?

jackpot51 (approved):

PR #645, opened 2024-11-05
Update cosmic_files.ftl

bittin:
Update Swedish translation


jackpot51 (approved):

bittin:
it works, just tested, will continue with cosmic-settings when i have less fever and cold

PR #649, opened 2024-11-11
fix(redox): Add "process" feature to tokio

joshuamegnauth54:
I'm going to be totally honest. I have no clue why this is compiling on Arch and Pop! _without_ the `process` feature on Tokio. However, it's failing on Redox as of today because my last patch uses [tokio::process](https://docs.rs/tokio/latest/tokio/process/) which is feature gated.

Anyway, adding process fixes building on Redox. :man_shrugging:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,7 @@ regex = "1"
 serde = { version = "1", features = ["serde_derive"] }
 shlex = { version = "1.3" }
 tempfile = "3"
-tokio = { version = "1", features = ["sync"] }
+tokio = { version = "1", features = ["process", "sync"] }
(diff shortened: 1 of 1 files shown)


git-f0x:
This seems to have been added in the meantime. :)

joshuamegnauth54:
Oh nice! I should close this then. :smiley_cat: 

I'm not sure why it worked without `process`. I remember I had a similar issue a while ago in reverse. One of COSMIC's projects repeatedly failed to compile when I tried it, but it was clearly passing CI and working for everyone else. It resolved itself within a day from someone else's patch. Very weird!

Anyway, thanks for letting me know. :grin:

PR #652, opened 2024-11-12
Fix applet window styling

git-f0x:
Fixes the scrollbar placement of `cosmic-files-applet` windows to match the planned design for other windows (the scrollbar should be at the edge of the window).
Also updates the icon to the latest one from the designs.

Initially wanted to wait for pop-os/libcosmic#706 and move stuff to the context drawer header, but updating libcosmic currently causes the menu buttons to be unclickable.


git-f0x:
This doesn't really need to be a draft anymore, since the moving the footer `layer_container` creation to libcosmic interferes with some possible layouts, so no changes are required here.

git-f0x:
The updated context drawers have some issues, mainly that the file picker details don't have the header actions, since I wasn't sure how to do it.
The preview/show details content also has extra padding for some reason, even without the header change.
The existing code can probably be cleaned up and refactored to better work with header actions in context drawers.

git-f0x:
This should now be fully functional. Still not sure where the extra padding of the previews is coming from, but that's just a minor visual issue.

Issue #332, opened 2024-08-09
Add size_menu entry to Folders Info


Moon-Mind:
![screenshot-2024-08-09-14-50-49](https://github.com/user-attachments/assets/4b580b9c-e4f4-4c78-a3a2-46fed1b4ab35)

PR #656, opened 2024-11-14
Directories sizes in details view

francesco-gaglione:
Folder size in details view:
image

### DIFF
--- a/src/tab.rs
+++ b/src/tab.rs
@@ -492,7 +492,7 @@ pub fn item_from_entry(
         selected: false,
         highlighted: false,
         overlaps_drag_rect: false,
-        size: 0,
+        size: None,
     }
 }
 
@@ -722,7 +722,7 @@ pub fn scan_trash(sizes: IconSizes) -> Vec<Item> {
                     selected: false,
                     highlighted: false,
                     overlaps_drag_rect: false,
-                    size: 0,
+                    size: None,
                 });
             }
         }
@@ -906,7 +906,7 @@ pub fn scan_desktop(
             selected: false,
             highlighted: false,
             overlaps_drag_rect: false,
-            size: 0,
+            size: None,
         })
     }
 
@@ -1310,7 +1310,7 @@ pub struct Item {
     pub selected: bool,
     pub highlighted: bool,
     pub overlaps_drag_rect: bool,
-    pub size: u64,
+    pub size: Option<u64>,
 }
 
 impl Item {
@@ -1410,9 +1410,11 @@ impl Item {
             ItemMetadata::Path { metadata, children } => {
                 if metadata.is_dir() {
                     details = details.push(widget::text(fl!("items", items = children)));
-                    let size = self.size;
-                    details =
-                        details.push(widget::text(fl!("item-size", size = format_size(size))));
+                    let size = match self.size {
+                        Some(size) => format_size(size),
+                        None => fl!("calculating"),
+                    };
+                    details = details.push(widget::text(fl!("item-size", size = size)));
                 } else {
                     details = details.push(widget::text(fl!(
                         "item-size",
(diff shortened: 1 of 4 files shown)


francesco-gaglione:
Close #332

src/tab.rs
```diff
@@ -1394,33 +1394,41 @@ impl Item {
...
+        details = details.push(widget::text(fl!("type", mime = self.mime.to_string())));
         let mut settings = Vec::new();
-        //TODO: translate!
-        //TODO: correct display of folder size?
         match &self.metadata {
             ItemMetadata::Path { metadata, children } => {
                 if metadata.is_dir() {
-                    details = details.push(widget::text(format!("Items: {}", children)));
+                    details = details.push(widget::text(fl!("items", items = children)));
+                    let size = calculate_dir_size(self.path_opt().unwrap());
```
jackpot51, line 1403: This will run every time the view needs to be remade (after every message) and will block rendering until it is complete. Please comsider using a subscription to load this information in the background

jackpot51 (changes requested):

src/tab.rs
francesco-gaglione: sure, I'm not very familiar with subscription in libcosmic, used once. Do you have a working example similar to this case?

src/tab.rs
```diff
@@ -1634,6 +1649,25 @@ pub struct Tab {
...
+fn calculate_dir_size(path: &Path) -> u64 {
+    let mut total_size = 0;
+
+    if let Ok(entries) = fs::read_dir(path) {
+        for entry in entries.flatten() {
+            let metadata = entry.metadata().ok();
+
+            if let Some(data) = metadata {
+                if data.is_dir() {
+                    total_size += calculate_dir_size(&entry.path());
```
jackpot51, line 1661: Recursion is problematic as it can cause stack overflows. Consider using walkdir::WalkDir instead.

src/tab.rs
```diff
@@ -1301,6 +1307,7 @@ pub struct Item {
     pub selected: bool,
     pub highlighted: bool,
     pub overlaps_drag_rect: bool,
+    pub size: u64,
```
jackpot51, line 1310: Consider using Option<u64> here so it can be determined if the size is still calculating or it is actually 0.

src/tab.rs
```diff
@@ -4411,7 +4451,35 @@ impl Tab {
...
                         Ok(()) => {}
                         Err(err) => {
-                            log::warn!("failed to send thumbnail for {:?}: {}", path, err);
+                            log::warn!("failed to send thumbnail for {:?}: {}", &path, err);
+                        }
+                    }
+
+                    let mut total_size = 0;
+
+                    if let Ok(entries) = fs::read_dir(path.clone()) {
```
jackpot51, line 4460: Now that you are using walkdir you don't have to do this loop any more

francesco-gaglione: Ops, sorry, fixed it

src/tab.rs
```diff
@@ -4411,7 +4451,19 @@ impl Tab {
                     match output.send(message).await {
                         Ok(()) => {}
                         Err(err) => {
-                            log::warn!("failed to send thumbnail for {:?}: {}", path, err);
+                            log::warn!("failed to send thumbnail for {:?}: {}", &path, err);
+                        }
+                    }
+
+                    let total_size = calculate_dir_size(&path);
```
jackpot51, line 4458: I think you need a new subscription for this, it looks like this only runs when thumbnailing a file?

jackpot51: Do you mind if I make some changes as I figure things out?

francesco-gaglione: I initially tried to create a dedicated subscription but had a number of problems with ownership which were later solved. However, I realised that every time this subscription is used, the size of the directory can be calculated. This subscription populates changes in the dialogue just where the size of the directory is to be shown, so in a way they are related. If that doesn't add up, I can try creating a separate subscription.

jackpot51: Please see this commit: https://github.com/pop-os/cosmic-files/pull/656/commits/f1c535f5f7a322cab969d8623dd6def0d2d30b88

jackpot51 (approved):

PR #659, opened 2024-11-17
feat: Navigate by typing file/directory names

sandr01d:
This PR adds the ability to jump to a file or directory by typing the first letters of its display name. Navigation will jump to the next matching item based on the current selection. The same character can also be typed in quick succession to cycle through all items beginning with the same character.
The typed word is reset when no letter has been typed for longer than defined in `PrefixSearch::reset_delay`. Currently this value is hard coded to 500ms, but could be easily made configurable. I did not do so yet, since this involves adding another PR for the settings and I wanted to get feedback for this one first.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -471,6 +471,49 @@ pub struct FavoriteIndex(usize);
 
 pub struct MounterData(MounterKey, MounterItem);
 
+struct PrefixSearch {
+    search: String,
+    reset_delay: u64,
+    next_reset: Option<Instant>,
+}
+
+impl PrefixSearch {
+    pub fn new(reset_delay: u64) -> PrefixSearch {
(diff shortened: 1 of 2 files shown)


jacobgkau (changes requested):
I think the majority of this PR was obsoleted by https://github.com/pop-os/cosmic-files/pull/1471. (Your PR was opened much earlier, at a time when we were focused on other functionality and weren't sure if this feature would be acceptable UX-wise; sorry about that.)

Repeating a key to cycle through items starting with that letter would be a welcome addition in a new or rebased PR.

sandr01d:
Thanks for the update. I'll look into the key cycling feature. Given that most of the changes in this PR are obsolete I'll open a new PR once I'm ready and close this one.

PR #675, opened 2024-12-03
feat: avoid overlap with panel

wash2:


src/app.rs
```diff
@@ -3303,6 +3364,45 @@ impl Application for App {
...
             Message::None => {}
+            #[cfg(all(feature = "desktop", feature = "wayland"))]
+            Message::Overlap(overlap_notify_event) => match overlap_notify_event {
+                OverlapNotifyEvent::OverlapLayerAdd {
+                    identifier,
+                    namespace,
+                    logical_rect,
+                    ..
+                } => {
+                    if namespace == "Dock" {
```
Drakulix, line 3375: Do we have to filter by namespace? Every exclusive layer-surface would be problematic here, no?

wash2: This is intended to avoid overlap with the auto-hide panel and dock, which don't have an exclusive zone. When they are exclusive, the surface should already not overlap.

Drakulix: Only, if they are on the same layer, which another panel like waybar and cosmic-files wouldn't be.

If this is for our auto-hide functionality, that makes sense, but I would like the PR to also include any exclusive surfaces on the top-layer, even without namespace matches. Could we do that?

wash2: Sure :+1:

Drakulix (approved):
LGTM! Thanks :)

src/app.rs
ids1024: Yeah. I would assume if any other client has an exclusive layer shell region (such as a dock or panel other than cosmic-panel), it should have the same behavior.

Issue #674, opened 2024-12-02
Unset "Show details" when the ContextDrawer is closed with the "Close >" button


WatchMkr:
Otherwise, cosmic-files unintentionally starts with ContextDrawer open. The setting is set/unset correctly when using kb shortcuts.

PR #678, opened 2024-12-04
Unset `show_details` on escape

joshuamegnauth54:
Closes: #674

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -326,6 +326,7 @@ pub enum Message {
     SearchActivate,
     SearchClear,
     SearchInput(String),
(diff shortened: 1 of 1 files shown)


jackpot51 (approved):

git-f0x:
I don't think this fixes the issue. It does work when pressing Escape, but it doesn't work when pressing the Close button in the context drawer (the show details config doesn't change).

joshuamegnauth54:
@git-f0x I didn't notice that! :sweat_smile: I usually use keyboard shortcuts so I'll check how to fix it for the Close button too.

PR #683, opened 2024-12-07
Add 'Single click to open' settings toggle

### LINKED ISSUE #330: Single-Click to Open
I always use Nautilus and Dolphin with Single-Click open because it's faster than clicking twice.

Nautilus' implementation leaves a little to be desired because it doesn't introduce handles in the top right you can click to select the file/folder when in Single-Click mode, so you need to `CTRL+CLICK` to select files/folders. Dolphin's behaviour is better here. But both implementations work fine.

CaelusV:
Fixes #330

### DIFF
--- a/i18n/en/cosmic_files.ftl
+++ b/i18n/en/cosmic_files.ftl
@@ -211,6 +211,7 @@ calculating = Calculating...
 
 ## Settings
 settings = Settings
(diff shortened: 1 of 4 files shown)


jackpot51:
This needs to be rebased

jackpot51 (approved):

PR #711, opened 2024-12-31
fix: update search on events

### LINKED ISSUE #281: Search results > Can't rename

ToBinio:
resolves #281 - [comment](https://github.com/pop-os/cosmic-files/issues/281#issuecomment-2423350968)

This now checks if a tab currently in search mode and updates search instead of rescanning the tab when an event happens.

It also fixes a bug where when in search some event e.g. rename from another program happens, the search would get cleared.

---

This might make:

https://github.com/pop-os/cosmic-files/blob/3dde3c8fbc5d92bcbfa61f53cb60d14b9b10f1cc/src/app.rs#L2540-L2541

Obsolete, based on my testing it does, but I am not 100% sure.

---

I also thought about introducing a new function that handles the logic to decide if a rescan or a search update needs to happen, and use that function everywhere instead of the rescan. I think this should be a bit more stable than handle each search edge case separate if more come up.

If this sounds reasonable, let me know.  

---

This also includes formatting fix for unformatted code introduced in #710 sry...

If there is anything I can make better in terms of PR's,
Please let me know so I can make your life as easy as possible :)

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -879,7 +879,7 @@ impl App {
 
     fn search(&mut self) -> Task<Message> {
         if let Some(term) = self.search_get() {
-            self.search_set(Some(term.to_string()))
+            self.search_set_active(Some(term.to_string()))
         } else {
             Task::none()
         }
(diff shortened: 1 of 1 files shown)


jackpot51 (approved):

jackpot51:
> I also thought about introducing a new function that handles the logic to decide if a rescan or a search update needs to happen, and use that function everywhere instead of the rescan. I think this should be a bit more stable than handle each search edge case separate if more come up.

Yes, this sounds good to me.

PR #735, opened 2025-01-11
Context menu fixes

### LINKED ISSUE #493: [BUG] Context Menu Not Dismissing on Click in Empty Directory's Navigation Area
**Cosmic-files version:**
```
cosmic-files:
  Installed: 0.1.0~1727141419~22.04~e5674da
```

**Issue/Bug description:**
In Cosmic Files, the context menu does not dismiss when clicking on an empty space in the navigation area of an empty directory. However, this behavior works correctly when the directory contains files or folders.

**Steps to reproduce:**
1. Open Cosmic Files
2. Open a empty directory
3. Right click on empty area on navigation area to open context menu
4.  Try dismissing the context menu by clicking on empty navigation area

**Expected behavior:**
The context menu should dismiss when clicking on any empty area in the navigation space, regardless of whether the directory contains files or not.

**Other notes:**
Similar bug has been reported here #360

jasonrhansen:
Adds the following behaviors:
-  Clicking in an empty folder clears the context menu. This also works for desktop mode when there are no icons.
- Clicking out of the COSMIC Files window clears the context menu. I was hoping this would also work for desktop mode so that clicking into an application window would clear the desktop context menu, but it doesn't look like it gets a window unfocused event.

Closes #493

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -352,6 +352,7 @@ pub enum Message {
     WindowClose,
     WindowCloseRequested(window::Id),
     WindowNew,
+    WindowUnfocus,
     ZoomDefault(Option<Entity>),
     ZoomIn(Option<Entity>),
     ZoomOut(Option<Entity>),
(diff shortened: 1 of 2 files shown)


jackpot51 (approved):

jackpot51:
The desktop mode probably gets a layer shell surface event for unfocus.

PR #777, opened 2025-01-29
chore(clippy): `tab.rs`

joshuamegnauth54:
As of this PR, Clippy only raises three warnings. One warning is related to a TODO which I mentioned in the last PR. Another warning involves an unused variable when a feature is disabled.

### DIFF
--- a/src/tab.rs
+++ b/src/tab.rs
@@ -83,8 +83,8 @@ const MAX_SEARCH_RESULTS: usize = 200;
 const THUMBNAIL_SIZE: u32 = (ICON_SIZE_GRID as u32) * (ICON_SCALE_MAX as u32);
 
 //TODO: adjust for locales?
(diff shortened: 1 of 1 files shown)


src/tab.rs
```diff
@@ -4721,6 +4684,200 @@ pub fn respond_to_scroll_direction(delta: ScrollDelta, modifiers: Modifiers) ->
     None
 }
 
+#[derive(Clone)]
+pub struct ArcElementWrapper<M>(pub Arc<Mutex<Element<'static, M>>>);
```
joshuamegnauth54, line 4688: The third unfixed lint is a warning for this line: [arc_with_non_send_sync](https://rust-lang.github.io/rust-clippy/master/index.html#arc_with_non_send_sync)

I don't think `ArcElementWrapper` is used across threads or else Rust wouldn't compile it. It's likely fine to use an `Rc<RefMut<...>>` instead but I left it as is.

jackpot51 (approved):

Issue #468, opened 2024-09-17
Compress/extract password protected ZIP files


jackpot51:
Related to #122

jackpot51:
Moving to alpha 3.

iamkartiknayak:
Would you implement support for `.7z` format as well?

jackpot51:
Support for other archive types should be in a new issue, not here.

PR #779, opened 2025-01-29
adding in support to extract/compress zip files with passwords

ellieplayswow:
apologies in advance if this is bad/jumping the gun/etc, i'm working for a client who _loves_ sending password protected zips atm, so this was a frequent pain point for me

**changes tl;dr:**
- refactor `operation::Operation::perform` to return `Result<OperationSelection, OperationError>` to bubble up specific errors
- 99% of existing errors bubble up as a `OperationErrorType::Generic(String)`, and implements the `std::fmt::Display` trait, so should be compatible across the board
- optionally ask for password when extracting a zip provides an invalid password error
- optionally ask for password when compressing a zipfile

--

 **when extracting:**
- first try to extract normally, however if an `OperationError` with `kind = OperationErrorType::PasswordRequired` is captured, open a new dialog prompting for password
- completing the new dialog tries to extract with the typed password
- if typed password incorrect, prompt again

**NB:** when a zip file has no password protection, functionality remains the same

--

**when compressing:**
- when zip is selected (by default) show a new password field
- when password is set, compress zip in `Aes256` mode with password prompted

**NB:** when no password is set, functionality remains the same

### DIFF
--- a/i18n/en/cosmic_files.ftl
+++ b/i18n/en/cosmic_files.ftl
@@ -39,6 +39,9 @@ resume = Resume
 ## Compress Dialog
 create-archive = Create archive
 
+## Extract Dialog
+extract-password-required = Password required
+
 ## Empty Trash Dialog
 empty-trash = Empty trash
 empty-trash-warning = Are you sure you want to permanently delete all the items in Trash?
--- a/src/app.rs
+++ b/src/app.rs
@@ -71,6 +71,7 @@ use crate::{
     spawn_detached::spawn_detached,
     tab::{self, HeadingOptions, ItemMetadata, Location, Tab, HOVER_DURATION},
 };
+use crate::operation::{OperationError, OperationErrorType};
 
 #[derive(Clone, Debug)]
 pub enum Mode {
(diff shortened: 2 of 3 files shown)


ellieplayswow:
should resolve #468

jackpot51 (changes requested):
Thanks for this, it is almost good for merge. There is an issue when extracting a password protected zip file that it will create the directory where it wants to extract it, then after putting in the password it will create another directory with `(Copy 1)` appended.

ellieplayswow:
@jackpot51 this is now resolved, but i'm not totally sure if my idea of solving this is the best.

tl;dr - the zip standard doesn't have the folder structure as encrypted, so `entry 0` can be a directory. if it is, we try and create the dir (sane!) and then skip to the next entry

instead i've updated it to push into a `VecDeque` (to maintain the order), and when we create a file instead of a dir we first pop out all of the queued entries (basically just delaying the directory creation until we know we can [un]successfully extract a file)

ellieplayswow:
cleaned up a few weird cases - couldnt exit a 'password required' dialog, allowing strings to be translatable - seems OK from my end

jackpot51 (approved):

Issue #641, opened 2024-11-04
[BUG?]: High memory consumption - spawn too many system threads for image/thumbnail decoding?


l-const:
**Cosmic-files version:**

20f9292

**Issue/Bug description:**

**Steps to reproduce:**

**Expected behavior:**

**Other notes:**

using heaptrack: https://github.com/KDE/heaptrack

installed with : 
```bash
dnf install heaptrack
```

executed:
```bash
heaptrack --pid ${pid of cosmic-files}
```

with initial on startup memory consumption of 48MB as seen from this:

![screenshot-2024-11-04-21-17-30](https://github.com/user-attachments/assets/b96a69b4-e81c-4fc4-9733-8700d1a4b882)

reaching up to and excedding 300MB:

![screenshot-2024-11-04-21-56-39](https://github.com/user-attachments/assets/6e6443f7-6fc1-4560-a883-d36a5b30044e)

heaptrack overview: 

![screenshot-2024-11-04-21-29-27](https://github.com/user-attachments/assets/dc5a0c55-052f-4a64-83bf-9b139b9fd68b)

![screenshot-2024-11-04-21-44-31](https://github.com/user-attachments/assets/64a57dd6-7f7c-4419-a15a-617ea07475a7)

it seems that no memory leak is happening but for the thumnail generation
we create through tokio too manu system threads by calling the clone3 system call

![screenshot-2024-11-04-22-02-32](https://github.com/user-attachments/assets/921f28c1-e0ea-44cb-b5fe-35d35a9c9438)

Here is the zipped .zst file that i exported:

[heaptrack.cosmic-files.31272.zip](https://github.com/user-attachments/files/17624124/heaptrack.cosmic-files.31272.zip)

l-const:
Used `top` to find the number of threads from the process, there is a variable size of threads , starting with 31 and managed to get up to 43 but then again did go down to 31 again after inactivity, the worrying part is that i managed to reach 1.0 GB memory consumption:

executed:

```bash
top -H -p 6880
```

![screenshot-2024-11-05-10-54-24](https://github.com/user-attachments/assets/4315da40-bd43-4ac9-b74d-d36a4707a661)
![screenshot-2024-11-05-10-56-54](https://github.com/user-attachments/assets/eaafa144-84b6-43ba-9c0c-1869bf7c6cb7)

l-const:
Update: 
1)  reached 1.2 GB
[...]

PR #789, opened 2025-02-05
fix: set M_MMAP_THRESHOLD to lower memory usage significantly

mmstick:
Similar to https://github.com/pop-os/cosmic-bg

Reduced memory usage on my system from 300+ MB to 40 MB after navigating through a couple directories with images. Allows most memory to be reclaimed when navigating between directories.


l-const:
thank you @mmstick  , i will proactively close this: https://github.com/pop-os/cosmic-files/issues/641

jackpot51 (approved):

Issue #568, opened 2024-10-04
[FEATURE REQUEST] Enable Installed App Selection for "Open With" Option in Cosmic Files


iamkartiknayak:
**Cosmic-files version:**
```
cosmic-files:
  Installed: 0.1.0~1727993042~24.04~d3f91e6
```

**Issue/Bug description:**
Currently, when right-clicking on a file/dir and selecting the "Open With" option, it does not display all installed applications that can handle the file type.

**Current behavior:**

![image](https://github.com/user-attachments/assets/52fabc58-bd8d-462f-86b2-01c6c8ad7861)

**Expected behavior:**
Something like this, where the first col would be recommended & others would also follow a MIME hierarchy based on if app supports the format. 
Dark tint for each item when hovered

![open_with_dialog_design_2024-10-04_200038_7948](https://github.com/user-attachments/assets/89ca8533-17bd-49f5-9dac-97c6ec3821df)

**Other notes:**

PR #792, opened 2025-02-05
Adding in new functionality to open / open-with MIME subclasses

### LINKED ISSUE #639: [bug] The text/* mime-type handler is a bit too strict
**Cosmic-files version:**
`0.1.0~1730233777~24.04~20f9292`

**Issue/Bug description:**
- The mime-type file-opener really strictly handles its types;
- It even does so when using "open with";
- I cannot choose an "other application" to try to handle a file, it only allows me to download a flatpak which also handles the exact same filetype;
- For example, when opening a `*.tsv` file, it correctly detects `text/tab-separated-values`, but it only allows me to use libreoffice calc.

**Steps to reproduce:**
- Create a file `/home/tom/somefile.tsv`
- Navigate to `/home/tom` in cosmic-files
- Right-click on `somefile.tsv`
- Click on "Open with..."

**Expected behavior:**
I would expect that it would allow me to use any application (that handles plaintext, `text/plain`, `text/*` for that matter) to edit a plain text file such as `.tsv`. For example, neovim [handles text/plain](https://github.com/neovim/neovim/blob/master/runtime/nvim.desktop), but I cannot in any way use Neovim to edit a `tsv` file through cosmic-files.

**Other notes:**
Nautilus does allow the user to specify any text editor (or any application at all for that matter, but full disclosure I don't like that design) to open `text/tab-separated-values`:
screenshot-2024-11-04-13-46-26

ellieplayswow:
~~this is a draft PR that requires https://github.com/ebassi/xdg-mime-rs/pull/31 to be merged upstream first~~

**changes tl;dr**
- allow opening / open-with based off of MIME subclasses (eg: handlers for `text/plain` for a JSON or CSV file)
- **when opening**: if a file has no direct handler, fall back to the first available subclass handler BEFORE using `open` crate
- **when open-with'ing**: add subclass handlers to the list of available programs below any exact handlers

demonstration vid:

https://github.com/user-attachments/assets/4a0d2b18-6ef3-41cd-bfea-2e2d378d0e93

(LHS defaults to `open` crate since i have no exact handler for `application/toml`)

fixes #639 , #568 (UI is slightly different)


ellieplayswow:
did a bit of cleanup and just yolo included my branch for `xdg-mime` once it gets merged upstream i'll monitor & remove the patch, or ideally the  `pop-os` org can host it

jackpot51 (approved):

Issue #190, opened 2024-05-25
 Drag-selecting does not scroll content


WatchMkr:
Need to be able to drag-select above and below the view area.

Same issue in cosmic-edit https://github.com/pop-os/cosmic-edit/issues/154

PR #812, opened 2025-02-18
adding in drag-scroll functionality when cursor is outside of scrollable viewport

ellieplayswow:
fixes https://github.com/pop-os/cosmic-files/issues/190

**changes tl;dr**:
- when dragging to select a number of items, mouse movements outside of the of the `Scrollable`/`MouseArea` viewport aren't captured
- we now capture `Event::Mouse(CursorMoved)` in `app.rs` to track global cursor position
- we bubble that event up to the active tab, which will scroll based on where the cursor is relative to the `Scrollable` viewport

this PR works and _could_ be merged into upstream, but it's currently a bit finicky:
- items that would be in the new selection don't get highlighted until you re-enter the viewport of the `Scrollable`;
- the speed is static (set to an arbitrary `5px`), it would be nicer if it was easier to control based off of how close your cursor is to the viewport
(i'll get these sorted in the next day or 2, i'm just pushing this up as it's late and otherwise am happy)

demonstration vid:

https://github.com/user-attachments/assets/6e9512b8-746d-4e06-b587-b927fe711dba

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -18,6 +18,7 @@ use cosmic::{
     app::{self, context_drawer, message, Core, Task},
     cosmic_config, cosmic_theme, executor,
     iced::{
+        self,
         clipboard::dnd::DndAction,
         event,
         futures::{self, SinkExt},
@@ -332,6 +333,7 @@ pub enum Message {
     Rename(Option<Entity>),
     ReplaceResult(ReplaceResult),
     RestoreFromTrash(Option<Entity>),
+    ScrollTab(f32),
     SearchActivate,
     SearchClear,
     SearchInput(String),
@@ -550,6 +552,7 @@ pub struct App {
     tab_dnd_hover: Option<(Entity, Instant)>,
     nav_drag_id: DragId,
     tab_drag_id: DragId,
+    auto_scroll_speed: Option<f32>
 }
 
 impl App {
@@ -1628,6 +1631,7 @@ impl Application for App {
             tab_dnd_hover: None,
             nav_drag_id: DragId::new(),
             tab_drag_id: DragId::new(),
+            auto_scroll_speed: None,
         };
 
         let mut commands = vec![app.update_config()];
@@ -2794,6 +2798,10 @@ impl Application for App {
                     self.operation(Operation::Restore { items: trash_items });
                 }
             }
+            Message::ScrollTab(scroll_speed) => {
+                let entity = self.tab_model.active();
+                return self.update(Message::TabMessage(Some(entity), tab::Message::ScrollTab(scroll_speed)));
+            }
             Message::SearchActivate => {
                 return if self.search_get().is_none() {
                     self.search_set_active(Some(String::new()))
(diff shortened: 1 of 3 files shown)


ellieplayswow:
i'm generally happy with how this works now, but could do with a fresh set of eyes for a few things:
- scrolling is still linked to mouse movement, i'm not sure how best to decouple this?
- i have some performance concerns with sending so many individual scroll events, but maybe we fix that with a solution to the above point?

jackpot51:
You could take a look at how I added drag scroll in cosmic-edit, here: https://github.com/pop-os/cosmic-edit/commit/04f2812f4e548d360f471c0e659ef2b97e22341b

I used a timed subscription to emit an automatic scrolling event.

ellieplayswow:
@jackpot51 thank you, that's a great help - ill work on implementing that tomorrow

outside of that are the changes ok? feels pretty good to me, and moving to a timer will feel a lot better too

jackpot51:
@ellieplayswow yep, the changes look good otherwise

ellieplayswow:
@jackpot51 that works _except_ a few things which I'm not sure on:
- in dev profile, a panic occurs when autoscrolling due to https://github.com/pop-os/iced/blob/master/futures/src/subscription.rs#L249 (although, it does work in release mode) - we can't _not_ move, and we can't easily make `Message` derive the `Copy` trait, so there might not be an easy way to resolve this?
- setting the speed dynamically doesn't work, i'm not sure what happens under the hood but if i slap a `println!("{:?}", scroll_speed)` here: https://github.com/pop-os/cosmic-files/pull/812/files#diff-0b979b6de560b29c1d97336878db56179224aa21d96fe099630c6628479ed020R4731 it correctly recognises the up-to-date value, but it doesn't get passed into the subscription (i suspect there's some threadsafe reason for this, but i'm not good enough w/ rust to notice off the top of my head)

i wanted to raise these since i suspect that actually cosmic-edit will have the same / similar issues? i know that dev profile might not be a huge loss to you but, imo, breakpoints are a pretty nice feature to poke around unfamiliar code and a random panic might not be ideal?

other than that it should all be good, the scroll speed feels reasonable and even with the up to 3x it feels like a really good experience

ellieplayswow:
@jackpot51 i'm happy to implement whatever changes are necessary if that's preferred, some pointers/ideas/examples would just be handy is all - i don't want to be a burden to anyone and am just happy to contribute :D

jackpot51:
When dealing with this before it was resolved by calling the with function on the subscription that passes a value through as another argument to the callback: https://github.com/pop-os/cosmic-files/pull/726

ellieplayswow:
@jackpot51 that works for both points! unfortunately when the speed changes it seems to cause a hiccup in scrolling, so i've just left it as a static speed for the time being.

the other change is having the subscription take an `i16`. the logic i've settled on was that we convert from an `f32` to an `i16` by `floor(f32 * 10)`, which felt like a fine resolution, and then we'd just reverse that when an `f32` was needed for the actual scroll action. if this gets looked at in the future, IMO the resolution is important to ensure things feel smoother

.. this might be a bit overengineered though, so happy to just keep it an i16 across the board if that's easier

jackpot51:
@ellieplayswow that sounds good. Is there anything else you'd want to do or is this ready to merge?

ellieplayswow:
@jackpot51 i think it's good for now! it feels really responsive so as long as you guys are happy with it i think the speed can be kicked down the line as a nice-to-have for the future (if it's even desired)

would you like me to look at implementing similar as a fix for `cosmic-edit` so that works in dev profile too?

jackpot51:
Sure, that would help

PR #816, opened 2025-02-20
Bugfix to how virtual cursor is calculated before DragEnd when drag-scrolling

ellieplayswow:
**changes tl;dr**
- part of #812 involved adding a subscription to auto scroll
- doing this caused a desync where the virtual cursor was getting updated far more often than the subscription itself
- this meant, unless the cursor was moved back in bounds, when the `DragEnd` event is triggered the amount of selected files was significantly more than actually selected

**steps to replicate**
- open a folder with a lot of items
- begin drag-scrolling
- let go of your mouse when you're out of bounds

https://github.com/user-attachments/assets/241dcf85-0664-443d-b835-056020d8c503

sorry i didnt catch this earlier :(

### DIFF
--- a/src/mouse_area.rs
+++ b/src/mouse_area.rs
@@ -23,7 +23,6 @@ use cosmic::{
     widget::Id,
     Element, Renderer, Theme,
 };
-use cosmic::iced::window::RedrawRequest;
 use crate::tab::DOUBLE_CLICK_DURATION;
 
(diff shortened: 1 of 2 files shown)


ellieplayswow:
@jackpot51 ccing you in this - sorry for the hassle, i noticed this earlier in the day but didn't have a chance to fix til now

Issue #817, opened 2025-02-21
Bug Rendering Thumbnails


johnoye742:
There's a bug in the rendering of thumbnails that causes thumbnail icons usually those of images and documents that have preview with incompatible aspect ratios to have a weird alignment issue where the image aligns to the right, causing inconsistent behaviour. This is due to the way thumbnail images are rendered being ContentFit::Contain for the content_fit property of those icons. In the documentation of that enum it says if it is used on an image where the bounds and the content have incompatible aspect ratios it might have whitespaces on top/bottom or left/right which is what I think happened here. This inconsistency could potentially cause bigger problems when rendering thumbnails.

Examples of what I mean is below.

I have written a fix, which I will create a PR for and put the link to the PR in comments below, I hope my patch is accepted.

PS: this is one of my first issues on github so permit me if I'm doing something wrong

![Image](https://github.com/user-attachments/assets/df2578eb-4870-4ce0-aabc-a60d59700f84)
https://drive.google.com/file/d/1v9Rfp5K0iAJ0uHu6AJlBIDUkat6uCYvj/view?usp=sharing

git-f0x:
Thumbnails are also aligned to the bottom, so they're off-center in the highlight.
This is also an issue in some other apps, and started happening after the libcosmic rebase.

johnoye742:
So perhaps the math in libcosmic is the one to be fixed? Do you suggest
going on to fix it?

git-f0x:
Not sure if it's something that should be manually fixed, or fixed upstream.
E.g. in COSMIC Store, when shrinking the window horizontally on an app page, the images go down and leave a lot of empty space.
@jackpot51 any thoughts? (sorry for the ping)

jackpot51:
If other applications have this problem it seems like something libcosmic needs to fix

PR #818, opened 2025-02-21
Fixes issue #817 with thumbnails

johnoye742:
Fixes the issue with thumbnails by explicitly setting the icon widget width to always be the minimum preventing overflow.

Some before-after images are below.

Before:
image

After:
image

Before:
image

After:
image

I would like to say that there are some other extremes and even in most other ones that are not very visible, that bug is still present if you look close enough.


johnoye742:
The fix is cited above by github. I'd like to call on @jackpot51 to review this please :+1:

jackpot51 (approved):

PR #853, opened 2025-03-05
[Feature] settings option to change sidebar size

BanceDev:
The default sidebar size is quite wide. I added a settings option to choose between small, medium, and large presets. 

It seemed to me that with the way the UI was presently designed it would be a bit complicated to make the sidebar resizable. All I could find was using something like PaneGrid but that seemed like it would lead to a lot of refactoring.

If any changes to this design are ideal like something akin to PaneGrid or using a slider in the settings instead I would be willing to alter it. I just think this is a much needed feature.

https://github.com/user-attachments/assets/2d2ea09d-64b9-42c1-a841-f8be990abd3f

### DIFF
--- a/i18n/en/cosmic_files.ftl
+++ b/i18n/en/cosmic_files.ftl
@@ -245,6 +245,12 @@ match-desktop = Match desktop
 dark = Dark
 light = Light
 
+## NavBar Size
+navbar-size = Sidebar Size
+small = Small
+medium = Medium
+large = Large
+
 # Context menu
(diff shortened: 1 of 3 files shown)


iamkartiknayak:
Could the sidebar support drag-to-resize with three predefined steps, snapping to the nearest defined size while prioritizing the drag direction?

iamkartiknayak:
Also on small size it needs to support tool tip on hover. At least 1s delay would be great

Issue #484, opened 2024-09-21
[BUG]: Details Pane Doesn't Show Details for Multiple Files and Doesn't Close When No File is Selected


iamkartiknayak:
**Cosmic-files version:**
```
cosmic-files:
  Installed: 0.1.0~1726882610~22.04~73ed4cf
  Candidate: 0.1.0~1726882610~22.04~73ed4cf
```

**Issue/Bug description:**
- When selecting multiple files, the details pane does not show any combined details of selected files rather shows the details of last selected file or dir.
- When no files are selected, the details pane remains open empty instead of closing automatically.

**Steps to reproduce:**
- Open Cosmic Files
- Select multiple files with `Ctrl` + `Click` or select all with `Ctrl` + `A`
- Open details pane with `Space-bar` key

**Expected behavior:**
- When selecting multiple files, The details pane should either show combined information (e.g., total size, number of selected files) or clearly indicate that multiple files are selected.
- When no files are selected, The details pane should automatically close or display a message indicating that no files are selected.

**Other notes:**

https://github.com/user-attachments/assets/e5ae1ff7-6179-4f83-a38b-c2af44da70c6

EthraZa:
Also, there is times it is still coming already opened, like in browser (native Brave in my case) file picker, consuming the view space.

![screenshot-2024-10-18-15-14-54](https://github.com/user-attachments/assets/98f1f4b5-6172-40fe-bd6a-c34642d3598f)

usbeehu:
Also it would be nice if there would be a narrow bottom bar that would show the number of files within the folder and the total size out of the box without having to open anything. Alao maybe the amount of free disk space too. Like in Dolphin in KDE or Finder in macOS.

PR #857, opened 2025-03-06
feat: multiple selection preview page

thatdevsherry:
related to #484 

show total number of dirs and files selected in preview page when more than one item is selected.

Currently added an associated function in `Item` which takes in a list of selected items and creates a simple view that shows the number of files and directories selected.

I'm not sure if this is the right approach but opening PR for further discussion.

image


iamkartiknayak:
I think UI needs approval from the design team. Also it looks empty.

Nautilus multiple selection property dialog
Screenshot_2025-03-10_19-21-17

Issue #759, opened 2025-01-19
[Bug] Broken Navigation Button during Search


iamkartiknayak:
**Issue/Description:**
The back navigation button only works once, removing or adding a character based on last change. It wouldn't work until another character has been modified using backspace or other keys.

https://github.com/user-attachments/assets/a33cbf4d-084c-4971-b142-ab4a5869a69b

PR #859, opened 2025-03-07
Type to search or seek

jackpot51:
This provides the ability to have typing when no input is focused to directly open the search input (like Ctrl+F) or the location editing input (like Ctrl+L). The default is configurable in settings.

image


iamkartiknayak:
There's a related [bug](https://github.com/pop-os/cosmic-files/issues/759) related to search that might need fixing.

PR #865, opened 2025-03-08
Update favorites when dir is renamed or moved

### LINKED ISSUE #863: [Bug] Renaming a directory or file breaks its shortcut in the sidebar
**Issue/Description:**
When a directory or file is added to the sidebar (navbar) in COSMIC Files, renaming the original source causes the shortcut to break instead of updating to reflect the new name. This leads to broken or outdated sidebar entries.

**Version:**
```
cosmic-files:
  Installed: 0.1.0~1741363853~24.04~63ab36a
```

https://github.com/user-attachments/assets/2def16e3-3627-4832-b492-d43f1188e1ba

jasonrhansen:
Fixes #863

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -1567,8 +1567,8 @@ impl App {
         .into()
     }
 
-    // Update favorites based on a rename or a move.
-    fn update_favorites(&mut self, from: &Path, to: &Path) -> bool {
+    // Update favorites based on renaming or moving dirs.
+    fn update_favorites(&mut self, path_changes: &[(PathBuf, PathBuf)]) -> bool {
         let mut favorites_changed = false;
         let favorites = self
             .config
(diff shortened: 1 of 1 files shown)


iamkartiknayak:
Should this also address these instances where?
- items are renamed or moved via terminal or other file manager

Currently it only  works if done via COSMIC Files

jackpot51 (approved):

jasonrhansen:
@iamkartiknayak You are correct, this PR only handles the case where directories are renamed or moved within COSMIC Files. Modifications made externally are not handled.

I tested Gnome Files and it also only updates bookmarks when changes are made in Gnome Files itself. I haven't tried any other file managers.

iamkartiknayak:
I tried it in COSMIC Terminal and it didn't update. Need to try in different file managers

jasonrhansen:
@iamkartiknayak Right, with this PR, COSMIC Files will **not** update the favorites based on changes made externally, including with COSMIC Terminal or any other file manager.

What I was trying to point out with Gnome Files is that it also doesn't handle updating its bookmarks based on external changes. Renaming or moving within Gnome Files will update the bookmarks in Gnome Files, but changes made outside of it won't. This PR adds functionality to COSMIC Terminal similar to Gnome Files.

PR #869, opened 2025-03-10
Bugfix to open-with dialog to use a scrollable with a managed height

ellieplayswow:
Fixes a niche issue where if you have a lot of handlers for a file type the content goes off-screen

image

(taken from pop-os chat)

another example:

https://github.com/user-attachments/assets/878c5c80-0ba2-45e4-8233-108ff5792d24

(for the `256.0` magic number i just poked around with various numbers until it would seamlessly transition between full height & scrollable without impacting the buttons)

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -4290,7 +4290,10 @@ impl Application for App {
                 };
 
                 let mut column = widget::list_column();
-                for (i, app) in self.get_programs_for_mime(&mime).iter().enumerate() {
+                let available_programs = self.get_programs_for_mime(&mime);
+                let item_height = 32.0;
+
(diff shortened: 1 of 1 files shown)


ellieplayswow:
tried with different # of items, zooms, resolutions - all seems to be working, however at some fringe resolutions the scrollable 'snaps' to fill a larger space than it should

i _think_ this is actually some fringe iced / libcosmic issue where it thinks the shrink/fill can be larger than expected, but it doesn't impact the buttons and only happens for like 10px worth of real estate - tuning the 256.0 magic number just starts to make other cases fail

jackpot51 (approved):

PR #900, opened 2025-03-25
Add --version and -- help support with lexopt

LinuxBoy-96:
Like I did with cosmic-greeter (merged0, cosmic-player, cosmic-edit and cosmic-term:
This add support for --version and --help argument in CLI for extra information.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1260,6 +1260,7 @@ dependencies = [
  "icu_provider",
  "ignore",
  "image",
+ "lexopt",
  "libc",
  "libcosmic",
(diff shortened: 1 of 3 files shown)


jackpot51 (changes requested):
Similarly to the PR for cosmic-edit, this does not appropriately change how argument handling happens in other places

LinuxBoy-96:
> Similarly to the PR for cosmic-edit, this does not appropriately change how argument handling happens in other places

I'm clueless, this is supposed to be accessed from a terminal (like many other app) and it work great that way.

PR #911, opened 2025-03-31
feat: use io_uring / IOCP when available for async file IO

mmstick:
Spawns a single thread for handling async file IO on a [compio runtime](https://github.com/compio-rs/compio). It is a completion-based IO runtime that can dynamically select a polling mechanism at runtime. It defaults to io_uring on Linux, IOCP on Windows, and the polling crate everywhere else. On Linux systems where io_uring is unavailable or disabled, it will fall back to the polling crate.

This eliminates most of the threads that were needed previously. It significantly reduced the amount of memory needed in the recursive Context to get a good transfer rate for each copy operation—from a 4 MB buffer to 128 KB. Copies on a nvme drive are somewhat faster with the async IO changes, and use less CPU than before.

Although it uses a single thread, it still manages to 100% max out my nvme drive's activity for the whole duration of multiple long transfers running concurrently. But it would be possible to enable compio's dispatcher to spread operations across worker threads if necessary.

All but the extract and compress operations were updated to be async. I had to switch the `CondVar` in the `Controller` to a `tokio::sync::Notify` to prevent the IO thread from being put to sleep when an operation is paused. Fixed a deadlock in the `operation_copy` test function that was performing an operation without concurrently pulling from the channel in the operation. Reduced the rate that `Message::None` is sent from a subscription to trigger a UI redraw, and fixed it to not run when operations are paused. The operation method also now returns a streaming Task instead of using a subscription.


jackpot51 (approved):

src/operation/recursive.rs
```diff
@@ -242,31 +250,83 @@ impl Op {
...
+                            break;
+                        }
+                        Ok(count) => count,
+                        Err(why) => {
+                            ctx.buf = buf_out;
+                            return Err(why.into());
+                        }
+                    };
+
+                    let BufResult(result, buf_out) = to_file.write_at(buf_out, pos).await;
```
jackpot51, line 298: Is buf_out exactly count bytes?

jackpot51: Fixed in https://github.com/pop-os/cosmic-files/commit/d6f1efbf672ed400362736550e838a3e6508e71d

Issue #924, opened 2025-04-07
Open File/Open With context menu opens two instances of the app


cdtj:
Doubleclick/Open With... context menu opens two instances of the app, the first one has the correct file while second window displays "/home/user/%U can't be opened", I tried different combos for apps/file extensions, the behavior is the same for evince, firefox, chrome and pdf, png, jpg

`Version: 0.1.0~1743716711~24.04~38b98e8`

Not sure if this relevant in comments the behavior was mentioned here: #922 

Opening app through the console (such as: `evice ~/sample.pdf`) doesn't reproduce the behavior.

Tommimon:
If #930 has been merged, can we close this issue?

Issue #922, opened 2025-04-05
Broken MIME handling after #891


abhinavnatarajan:
Commit #891 seems to have broken MIME handling.

1) I can't seem to open files in flatpak apps anymore. For example, I have Okular installed as a flatpak app, with the following command in the desktop file:
   ```
   Exec=/usr/bin/flatpak run --branch=stable --arch=x86_64 --command=okular --file-forwarding org.kde.okular @@u %U @@
   ```
   Nothing happens when I double click on a PDF file in cosmic-files despite Okular being set as the default PDF viewer in cosmic-files via the details sidebar. Choosing Okular via left-click -> 'Open with...' also has the same effect. Directly running the above flatpak command with the filename in the terminal works as expected.
The same problem occurs with all other flatpak apps that I've tried with - all of the Libreoffice apps, gThumb, and Gnome Papers.

2) I'm using neovide (GUI frontend for neovim) as my default text editor, and the desktop entry simply has the line `Exec=neovide %F`. Double clicking on a text file opens the text file but also a new buffer called '%F'. Again, running `neovide <filename>` from a terminal works fine.

Notably the problem does not occur with cosmic apps, for example left-click -> 'Open With...' and choosing Cosmic Edit works correctly for text files. The problem also did not occur before version `0.1.0~1743716711~24.04~38b98e8`.

abhinavnatarajan:
On closer inspection of `mime_app.rs`, it seems like the current handling also could mess with the ordering of arguments specified in the `Exec` key of the desktop entry by always placing the file URI as the first argument. This can have potentially serious side-effects for programs that expect their arguments in a certain order or for desktop entries with more complicated commands.

Joomsy:
To add, the problem still exists for me when using the "Open with..." dialog, so it looks like this is application dependent. I've noticed this behavior with both file-roller and Sublime. Like with you and neovide, a secondary window will open, and it seems like arguments are being treated like files. For example, file-roller will say "Error opening file /home/$user/%U: No such file or directory".

abhinavnatarajan:
I had a chance to look at the code in more depth, and here's what seems to be happening: the arguments are iterated over twice, but only the substitution for the `%f`/`%F`/`%u`/`%U` argument is added to the command on the first pass. On the second pass, all the arguments are appended to the argument list (including the raw `%f`/... string). So the following `Exec` key
```
...
Exec=cmd arg1 arg2 %f arg3
...
```
will result in the following command being run 
```
[...]

Issue #925, opened 2025-04-07
[BUG] Some files associated with Flatpak Programs Won't Open (TeXStudio, Onlyoffice)


patrickkoval:
Hi,

I have flatpak versions of TeXStudio and Onlyoffice set for certain file types. These won't open from cosmic-files. I have to open the programs themselves to use the file picker from within the program.

Config below.

![Image](https://github.com/user-attachments/assets/fa9c3f5e-0bb7-49ec-aa53-fbe5b3067b07)

joshuamegnauth54:
#930 may fix this, but let me know if it doesn't.

patrickkoval:
I tried with both an OnlyOffice file and a TeXStudio file and neither flatpak version works.

This might be worthy of a different thread (don't know if you want to stick to one bug per thread) but to troubleshoot I uninstalled the flatpak TeXStudio and installed it from apt and was able to open a file. But when clicking on a particular file from cosmic-files, it opens two files: one the requested file and one labeled "%F". This also seems to happen with the document viewer where I'll open a PDF and it'll try to open a second file and throw an error saying that it can't find a file called "%U". I've attached both below.

![Image](https://github.com/user-attachments/assets/9fcdcefb-0a48-4c3a-b19b-f7a789494b33)

![Image](https://github.com/user-attachments/assets/fe54bc3a-4667-407d-a383-a4954e0d2a1d)

joshuamegnauth54:
Hi again! Are you still experiencing this problem? #930 should fix it, so if you still face issues after updating let me know.

patrickkoval:
Hi! Ah, that makes sense. I don't think it has, unfortunately. I tried both Onlyoffice and TeXStudio flatpaks with their respective files. Tried opening them regularly, "open with", and opening them when the program is already running to see if the file appears. Still doesn't seem like it's working.

The "%U"/"%F" file thing I mention above also seems to be happening.

Just to make sure I'm not doing something wrong, would these commits be available via the stating repositories?

jackpot51:
We will release the updates from staging sometime this morning

patrickkoval:
Looks like both issues I mentioned were resolved. Thanks!

PR #930, opened 2025-04-09
fix: Handle complex Exec keys in desktop entries

joshuamegnauth54:
Closes: #922, #925, #924

My original patch, #891, mishandled non-trivial Exec keys. The ordering of arguments matters for the keys, such as for Flatpak apps.


joshuamegnauth54:
This doesn't fix #924. I'll figure that out tomorrow, hopefully.

joshuamegnauth54:
If anyone notices any lingering issues from #891, open an issue and tag me. I'll try to handle it. I'm sure there are a few left that I haven't noticed in my day-to-day use.

jackpot51 (approved):

Issue #921, opened 2025-04-03
Fails to build on macOS


sophiajt:
I'm not entirely clear on the intent to support macOS, but I noticed that things like cosmic-text require this crate, and that feels a bit more general purpose than just cosmic.

Currently, when building on macOS, you get:

```
error[E0308]: mismatched types
    --> /Users/sophia/.cargo/git/checkouts/cosmic-files-1a0f61508ace3cc6/5732a8d/src/operation/mod.rs:1053:28
     |
1053 |                 return Err("Restoring from trash is not supported on macos".to_string());
     |                        --- ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `OperationError`, found `String`
     |                        |
     |                        arguments to this enum variant are incorrect
     |
help: the type constructed contains `std::string::String` due to the type of the argument passed
    --> /Users/sophia/.cargo/git/checkouts/cosmic-files-1a0f61508ace3cc6/5732a8d/src/operation/mod.rs:1053:24
     |
1053 |                 return Err("Restoring from trash is not supported on macos".to_string());
     |                        ^^^^------------------------------------------------------------^
     |                            |
     |                            this argument influences the type of `Err`
note: tuple variant defined here
    --> /Users/sophia/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/result.rs:537:5
     |
537  |     Err(#[stable(feature = "rust1", since = "1.0.0")] E),
     |
```

PR #942, opened 2025-04-20
Rudimentary continuous integration

joshuamegnauth54:
Related to #921

WIP


joshuamegnauth54:
Oh also I'll go on the record that YAML is way harder than Rust. :face_with_spiral_eyes:

jackpot51:
Is it possible to check this before merge?

joshuamegnauth54:
I'm not sure. I don't know much about CI/CD pipelines (they're harder than they look :sob: ). This PR is the result of going through GitHub's pages on pipelines and [this page](https://doc.rust-lang.org/cargo/guide/continuous-integration.html) from the Rust book.

I tested it with [Act](https://github.com/nektos/act), an open source program for running GitHub Actions locally. Linux works, but I couldn't test macOS and Windows because I only have Linux. I wrote the macOS pipeline based on research that mentioned that Homebrew can be used to install dependencies.

My hunch is that the Linux pipeline will still work fine if committed, but macOS and Windows will require some more work. Luckily, if macOS and Windows fail, there will be logs from GitHub showing what needs to be fixed. That's how I figured out to disable those tests (except I used Act).

.github/workflows/main.yml
```diff
@@ -0,0 +1,62 @@
...
+        uses: actions/checkout@v4
+
+      - name: Add Rust
+        uses: actions-rust-lang/setup-rust-toolchain@v1
+        
+      - name: Update ${{ matrix.name }}
+        run: |
+          rustup target add ${{ matrix.target }}
+          rustup update stable
+          rustup default stable
```
HigherOrderLogic, line 54:
```suggestion
      - name: Add Rust
        uses: actions-rust-lang/setup-rust-toolchain@v1
        with:
          toolchain: stable
          target: ${{ matrix.target }}
```

joshuamegnauth54: Thank you! Let me know if you have any more suggestions or improvements. I'll commit them, squash, and credit you.

.github/workflows/main.yml
```diff
@@ -0,0 +1,59 @@
...
+              --skip copy_file_to_same_location \
+              --skip copy_file_with_extension_to_same_loc \
+              --skip copy_file_with_diff_name_to_diff_dir \
+              --skip copy_to_diff_dir_doesnt_dupe_files \
+              --skip copying_files_multiple_times_to_same_location"
+
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
```
gwen-lg: It would probably be easier to understand if you put the strategy part first, and this line just before the line `steps`.

PR #1003, opened 2025-05-25
drastically speed up gvfs mount directory listings

ellieplayswow:
listing a directory inside a gvfs mount can be currently extremely laggy (ex: `DCIM/Camera` directory on my phone, which has ~1300 items, takes ~25 seconds to list everything)

**changes tl;dr**
- use `gio::prelude::FileExt::enumerate_children` to loop through a directory if it's inside a gvfs mount, rather than using `fs::read_dir`
- for gvfs directories, use a new `ItemMetadata::GvfsPath` to allow grabbing some meta directly from gio to avoid seemingly more expensive `fs::` operations
- for gvfs directories, always assume `remote = true` for the purposes of guessing mime types (to avoid having to read the file)

for me, this loads up my `DCIM/Camera` folder in ~75ms instead :D 

this is (mostly) functional, but leaving as a draft until i can fix the following:
- [x] thumbnail generation
- [x] show details
- [x] show image with `<SPACE>`

### DIFF
--- a/src/tab.rs
+++ b/src/tab.rs
@@ -587,12 +587,10 @@ pub fn parse_desktop_file(path: &Path) -> (Option<String>, Option<String>) {
 }
 
 #[cfg(feature = "gvfs")]
-pub fn item_from_gvfs_info(
-    path: PathBuf,
-    file_info: gio::FileInfo,
-    sizes: IconSizes,
-) -> Item {
-    let file_name = file_info.name().into_os_string().into_string().unwrap_or_default();
+pub fn item_from_gvfs_info(path: PathBuf, file_info: gio::FileInfo, sizes: IconSizes) -> Item {
+    let file_name = file_info
+        .attribute_as_string(gio::FILE_ATTRIBUTE_STANDARD_NAME)
+        .unwrap_or_default();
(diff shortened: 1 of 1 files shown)


ellieplayswow:
i'm happy with this overall now, things should remain unchanged for non-gvfs dirs. one thing i am noting is that thumbnail generation is _slow_ for my phone's camera folder, and in fact the whole app seems to crawl to a bit of a slow, but this isn't really new (i suspect that we may need to either reduce the number of thumbnailer jobs when in a gvfs dir or something else if it's not async?)

ps - it might be an idea to create a config for "read files to guess mime type", i remember seeing an issue someone reported where cosmic files would trigger clamav due to accessing every file, and having that as an optional (but default enabled) config flag may help fixing that

jackpot51 (approved):

ellieplayswow:
@jackpot51 i've resolved the merge conflict, however i've noticed that this version seems to behave differently (and i don't think due to my changes?)

specifically, if i visit my DCIM/camera folder, the whole application hangs for a few seconds before just displaying everything - i also had to add this commit: https://github.com/pop-os/cosmic-files/pull/1003/commits/3436e4fd31116a256353411af1c30c4da1c57bab which previously wasn't necessary. i feel like something extra is happening on the main thread?

PR #1042, opened 2025-06-27
add ability to pass appid through to file dialogs

ellieplayswow:
**changes tl;dr**
- add new `Dialog::create()` method w/ new `DialogSettings` struct
- deprecate `Dialog::new()` in favour of `Dialog::create()`

this allows passing the appid for file dialogs, eg if i'm trying to open a file/folder via an xdg portal it'll currently create a new window with the appid **com.system76.CosmicFilesDialog**. this allows the xdg portal to pass through the appid of the requester, meaning that the windows can stay stacked/associated.

this also provides a more backwards compatible way to add new settings to dialogs in the future

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -66,6 +66,7 @@ use trash::TrashItem;
 #[cfg(all(feature = "wayland", feature = "desktop-applet"))]
 use wayland_client::{protocol::wl_output::WlOutput, Proxy};
 
+use crate::dialog::DialogSettings;
 use crate::{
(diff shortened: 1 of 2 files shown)


ellieplayswow:
ah, i hadn't considered that, i wonder if there's an api or something to set a window's parent? since that's another arg that gets passed into the portal

jackpot51:
@ellieplayswow potentially we would need https://github.com/rust-windowing/winit/issues/159

PR #1046, opened 2025-06-30
Support CLI argument for Version and Help 

LinuxBoy-96:
This ensures that cosmic-term can provide it's version and some information via the terminal.
This is essential for when multiple distributions start using COSMIC and won't update at the same time.

This is a cleaner version, my previous attempt had way too much commit for a simple feature...

It use Clap_lex

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1032,6 +1032,12 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


mmstick:
This breaks argument parsing. You need to ensure that `--` is being handled properly.

Issue #858, opened 2025-03-07
Persist sort order per folder


WatchMkr:

johnoye742:
Thankfully someone finalky pointed this out.

johnoye742:
Any ideas where it's being set by default to A-Z, being new to the project it's still hard to tell the flow.

Could anyone point me in the direction of where in startup it is set, so I do not set it in a place where it'll be overriden.

WatchMkr:
Default downloads to Newest First.

PR #1048, opened 2025-07-01
feat: persist sort options for each location

wash2:
fixes https://github.com/pop-os/cosmic-files/issues/858


src/config.rs
```diff
@@ -110,6 +113,7 @@ pub struct Config {
     pub show_details: bool,
     pub tab: TabConfig,
     pub type_to_search: TypeToSearch,
+    pub sort_names: HashMap<String, (HeadingOptions, bool)>,
```
jackpot51, line 116: I believe this should be in state, not in config.

PR #1054, opened 2025-07-05
Add caching for thumbnails based and freedesktop.org spec

### LINKED ISSUE #707: [Enhancement] Cache support for thumbnails/album art
**Cosmic-files version:**
```
cosmic-files:
  Installed: 0.1.0~1734538799~24.04~ee7954e
```

**Issue/Bug description:**
Currently, there's no caching implemented in the Files app, causing it to use up to 40% of all CPU cores every time I reopen a directory full of images or PDFs. Implementing caching would significantly reduce CPU usage and greatly improve the user experience.

jasonrhansen:
Closes #707

Implements https://specifications.freedesktop.org/thumbnail-spec/latest/, which enables it to share thumbnails with other applications.

It doesn't handle cleaning up the thumbnail cache. It would probably be a good idea to have some mechanism to cleanup old cached thumbnails so the cache doesn't grow forever.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1507,6 +1507,7 @@ dependencies = [
  "libc",
  "libcosmic",
  "log",
+ "md-5",
  "mime_guess",
  "notify-debouncer-full",
  "notify-rust",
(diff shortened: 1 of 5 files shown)


jackpot51:
Amazing work, I will review on Monday

PR #1058, opened 2025-07-08
thumbnail: Support jxl and plumbing for future formats.

Quackdoc:
This will allow using jxl-oxide for thumbnail creation, it seems to be working well for me but it's a draft for a couple reasons, the primary one is im not sure the best resource limit to set. I feel like this should be set in a configuration, but even then I don't know what defaults should be used.

the current value I found to be necessary for some of my images, but likely way to high for lower end systems. (that being said, hitting an oom case in cosmic-files is much better then having a thumbnailer oom since oomd will actually kill cosmic-files instead of the entire compositor)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -171,6 +171,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "alloc-no-stdlib"
+version = "2.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cc7bb162ec39d46ab1ca8c77bf72e890535becd1751bb45f64c597edb4c8c6b3"
+
+[[package]]
+name = "alloc-stdlib"
+version = "0.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


jackpot51:
I would recommend a much lower memory limit and potentially a lower number of jobs.

Quackdoc:
I think this is more or less ready to review with a small cleanups I haven't pushed. should jxl-oxide try to convert the thumbnail to sRGB? It will help thumbnails look better, but may give false impression when an image is for instance an HDR image.

jackpot51:
Yes, the thumbnails should be sRGB.

Quackdoc:
ah, so jxl-oxide's image-rs doesn't currently handle the colorspace conversions, I can probably add it to jxl-oxide but that would probably put off the PR for a bit. 

Image-rs doesn't handle colorspace conversions anyways so the current configuration is inline with what it was doing before, I think using an external cms instead of exposing jxl-oxide's cms is probably better for consistency and can be done in a separate PR.

Issue #515, opened 2024-09-28
[BUG] Cosmic Files not showing image preview despite having thumbnailer support


iamkartiknayak:
**Cosmic-files version:**
```
cosmic-files:
  Installed: 0.1.0~1727456658~24.04~5e0ef57
```

**Issue/Bug description:**
Cosmic Files does not recognize certain image formats such as `.heic`, even after installing and configuring the appropriate thumbnailer support for those formats. It shows the thumbnail of image in navigation view but doesn't produce preview in preview pane. 

**Steps to reproduce:**

https://github.com/user-attachments/assets/960291c1-9245-47de-9ed6-6aa881c706d1

**Expected behavior:**
Thumbnails should be displayed in preview pane, provided that the necessary thumbnailer support is correctly installed.

**Other notes:**

Quackdoc:
I believe this is because cosmic-files uses `widget::image()` but does not manually create the image using thumbnailer. I am not particularly sure *how* this could work, cosmic could potentially rely on the thumbnailer for this considering that `%s` is optional, however in reality i am not sure how reliable this is. Perhaps cosmic-files could abstract out image decoding into a seperate function that could be shared across `widget::image()` and the thumbnailer

iamkartiknayak:
Still a issue in Alpha-6 for `.heic` image format

Quackdoc:
I am working on a PR to add support for jxl for the preview and gallery view, "In case if thumbnailer is not setup a generic image should be used in preview pane & Gallery view should not execute"

Is there an good image somewhere? I want to use a fallback image for when jxl fails to encode and figure it should wind up being the same one as this. 

as an additional question since I'm working on this anyways, is it guaranteed that a thumbnailer will produce acceptable results? I had at least one thumbnailer in the past force you to enter a value, and would always produce a square image. Would it be better to throw the generic image instead?

KodeBarista:
Now `.heic` files shows up in context-drawer but still no preview. 

<img height="400" alt="Image" src="https://github.com/user-attachments/assets/0d442a97-015d-4430-ab83-221f9f8d12b9" />

p-sims:
Relates to #510 
On Fedora Linux 43.20260903.0 it isn't consistent....
Some JXL file work, and for others it's like preview opens (window gets darker and arrow appear), but no image is display....

[...]

PR #1113, opened 2025-08-01
feat: jxl support for gallery and preview

Quackdoc:
Support JXL images in preview and thumbnail 

Things that need to be solved first 

* This is causing images not to render: https://github.com/tirr-c/jxl-oxide/issues/473
* I am using unwraps with the intention of having a place holder image as mentioned here https://github.com/pop-os/cosmic-files/issues/515

Questions

Should I also wire up using a thumbnailer for the preview pane and perhaps even gallery view? Personally I don't think gallery view should use the thumbnail but preview should.

Should the gallery view not open or should it display a place holder image? A placeholder image could show a "non compatible image" format which may be more user friendly then not opening the gallery at all.

Should I change thumbnailer to use a similar function (see below) or use the nested matches for gallery and preview?
```rs
jxl" => {
                    let jxldynimg = (|| {
                        let file = File::open(path).map_err(|err| {
                            log::warn!("failed to open file {:?}: {}", path, err);
                        }).ok()?;
                        let mut limits = image::Limits::default();
                        let max_ram = max_mem * 1000 * 1000 / jobs as u64;
                        limits.max_alloc = Some(max_ram);
                        let mut jxl = JxlDecoder::new(file).map_err(|err| {
                            log::warn!("to create JxlDecoder for {:?}: {}", path, err);
                        }).ok()?;
                        let _ = jxl.set_limits(limits);
                        Some(image::DynamicImage::from_decoder(jxl).map_err(|err| {
                            log::warn!("failed to decode the file {:?}: {}", path, err);
                            }).ok()?
                        )
                    })();
                    //TODO: This will be unwrap_or(generic image)
                    jxldynimg
                }
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2092,6 +2092,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "dunce"
+version = "1.0.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "92773504d58c093f6de2459af4af33faa518c13451eb8f2b5698ed3d36e7c813"
+
+[[package]]
 name = "dyn-clone"
 version = "1.0.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4322,6 +4328,7 @@ dependencies = [
  "jxl-oxide-common",
  "jxl-render",
  "jxl-threadpool",
+ "lcms2",
  "tracing",
 ]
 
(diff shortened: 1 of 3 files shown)


Quackdoc:
No real need to, if its better it can be swapped out, jxl-rs won't be ready for a while.

This is currently waiting on https://github.com/pop-os/libcosmic/issues/956

PR #1165, opened 2025-09-04
Migrate to Rust 2024 edition and update dependencies

### LINKED ISSUE #290: Icons in list view doesn't use bigger variants when scaling is increased
изображение

git-f0x:
The vergen crate removed git features, so I suppose it's a good time to switch to the `about` widget (closes #658).
Replaces `xz2` with `lzma-rust2`, since updating `zip` causes a conflict without that (and it's a pure Rust crate).

ICU isn't updated since the changes are very extensive, and I'm not yet sure how to adapt everything.

Edit: Also fixes #290 with the libcosmic changes of `icon::from_name` handling.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,8 +51,8 @@ xdg-mime = { git = "https://github.com/ellieplayswow/xdg-mime-rs", branch = "fea
 bzip2 = { version = "0.6", optional = true }           #TODO: replace with pure Rust crate
 flate2 = "1.1"
 tar = "0.4.44"
-liblzma = { version = "0.4.4", optional = true }       #TODO: replace with pure Rust crate
-ordermap = { version = "0.5.9", features = ["serde"] }
+lzma-rust2 = { version = "0.13", optional = true }
+ordermap = { version = "0.5.10", features = ["serde"] }
 # Internationalization
 i18n-embed = { version = "0.16", features = [
     "fluent-system",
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -44,48 +43,48 @@ tokio = { version = "1", features = ["process", "sync"] }
...
-png = "0.17.16"
-jxl-oxide = { version = "0.12.2", features = ["image"] }
+png = "0.18"
+jxl-oxide = { version = "0.12.4", features = ["image"] }
 num_cpus = "1.17.0"
 
 # Completion-based IO runtime to enable io_uring / IOCP file IO support.
 [dependencies.compio]
-version = "0.14.0"
+version = "0.15.0"
```
libkurisu, line 74: heyo. compio just dropped version 0.16.0 which fixes some issues with building on musl based distros.

wash2:
This seems to have somehow introduced an offset for the dnd cursor image. I'll help look into a fix for that, but it otherwise looks ok. I need to do some more testing to make sure the networks section is ok too.

mmstick:
Perhaps it'd be a good idea to expose the `-sys` features of these C libraries for the distributions that'd prefer to link to their system libraries?

wash2:
> > This seems to have somehow introduced an offset for the dnd cursor image.
> 
> I'm not noticing any offset different from before for dnd items. Is there any particular place to look?

Nevermind, I tested the wrong branch, looks good.

wash2 (approved):

PR #1184, opened 2025-09-16
Update cosmic_files.ftl ITA

lucamosca1:
Updated italian l18n

### DIFF
--- a/i18n/it/cosmic_files.ftl
+++ b/i18n/it/cosmic_files.ftl
@@ -5,7 +5,7 @@ no-results = Nessun risultato trovato
 filesystem = Filesystem
 home = Home
 networks = Reti
-notification-in-progress = Operazioni sui files in corso.
+notification-in-progress = Operazioni sui file in corso.
 trash = Cestino
 recents = Recenti
 undo = Annulla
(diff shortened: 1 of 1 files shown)


jsanchezba:
I would advice updating also these files so you can have the app name translated instead of being in english. (I think this is how they are translated)

- res/com.system76.CosmicFiles.desktop
- res/com.system76.CosmicFiles.metainfo.xml

i18n/it/cosmic_files.ftl
```diff
@@ -30,8 +30,14 @@ size = Dimensione
...
-operations-running = {$running} operazioni in corso: ({$percent}%)...
-operations-running-finished = {$running} operazioni in corso: ({$percent}%), {$finished} completate...
+operations-running = {$running} {$running ->
+    [one] operazione
+    *[other] operazioni
+  } running ({$percent}%)...
+operations-running-finished = {$running} {$running ->
+    [one] operazione
+    *[other] operazioni
+  } running ({$percent}%), {$finished} finished...
```
gabriele2000, line 40: "operazioni running finished" doesn't seem correct...

lucamosca1: you're right, going to fix it, thanks!

PR #1290, opened 2025-10-17
dialog focus

### LINKED ISSUE #983: Dialog not in focus after it appears
I was trying to delete a file by pressing "Shift+Del" and then pressing "Enter" to confirm the dialog, however this actually ended up opening the file instead - even though the confirmation dialog was visible, it didn't get focus and the folder behind it got focus instead, resulting in unusual behavior.

wash2:
fixes #983


jackpot51 (approved):

leviport (approved):
This is an improvement over current behavior, but I think there are still more improvements that could be made. Dialogs are now appearing with a button focused, so hitting Enter does work without tabbing or having to use the mouse. 

It doesn't make the dialog fully grab focus from the arrow keys though. So if I try to left-arrow over to the cancel button, the file selection underneath the dialog is changing, rather than the Cancel button gaining focus. Also, the button that started out focused when the dialog appeared loses focus. 

I also feel like there's precedent for having the Cancel button focused by default, at least for destructive actions like permanent deletion. That would make the user arrow to the affirmative button before hitting Enter, ensuring that it's a very deliberate action. I'm not certain of that though, so that could be wrong.

Another weird situation is the Open With... dialog. The Open button starts focused, so Enter opens the file in the app that's currently selected, but arrowing up and down does not change the app selection, and instead is changing the file selection in the cosmic-files window below the dialog.

Pinging @pop-os/ux so this is on their radar as well.

PR #1329, opened 2025-10-28
General code cleanup and performance optimisations

Cheong-Lau:
Some pedantic `clippy` fixes, code cleanups and other minor performance improvements. https://github.com/pop-os/cosmic-files/pull/1298#issuecomment-3419493807 has benchmarks of the older version of this PR, I expect more improvements to have been made since then. I can do more benchmarks if they're wanted, I just don't actually know what to test since these changes are so general?

Notably I've moved the code that handles zooming to its own module to deduplicate code, there's probably more that is very similar between `App` and `Dialog` but I'm not sure on how everything should be organised, so I've left it all for now.

Sorry for the large PR, but I have actually done a fair amount of testing this time and I think I haven't broken anything (?).

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -56,7 +56,6 @@ use std::{
     env, fmt, fs,
     future::Future,
     io,
-    num::NonZeroU16,
     path::{Path, PathBuf},
     pin::Pin,
     process,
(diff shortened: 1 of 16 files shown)


leviport (dismissed):
Testing complete:

### Basic navigation

- [ ] Hover over the top item in the folder, then scroll down until it's out of view (while still hovered).
      --> Hover highlight persists when hovered, scrolled out of view, and scrolled back to. Not a regression.
      --> Right clicking a sidebar item makes the fonts of all sidebar items turn bold. Clicking another sidebar item un-bolds all items except the one that was just clicked. Not a regression.

### File operations

- [x] Files can be renamed.

- [x] Image and video thumbnails display in local folders.
- [x] Sorting options work.

### External filesystems

There are two tests in "Basic navigation" that did not pass, but I confirmed that they are not regressions, and are present on current `master`

jackpot51 (dismissed):

jackpot51:
The conflicts with #1298 (now merged) must be resolved

Issue #1253, opened 2025-10-07
sshfs image viewing and thumbnail is broken


drwankingstein:
sshfs is a very nice way to view files on other devices but it seems to be explicitly blocked, please allow it.

PR #1224, opened 2025-09-29
Option to disable thumbnailing

joshuamegnauth54:
Closes: #1216

Issue #1216, opened 2025-09-27
[Feature Request] Add an option to disable thumbnailing


ghost:
### Petition
As of its current state, Cosmic Files doesn't provide an option to disable thumbnailing through the GUI provided (and I believe it isn't possible by editing the config files for Cosmic Files either). It would be great to have an option to completely turn off thumbnails through the GUI.

### Why is this needed?
[It is possible to attack users through thumbnailers](https://scarybeastsecurity.blogspot.com/2016/11/0day-exploit-compromising-linux-desktop.html). The ability to completely disable thumbnailing would not only help to prevent this type of exploits but would also help to increase performance on low-end machines.

There is already an existing [petition](https://github.com/pop-os/cosmic-files/issues/1189) to add thumbnailer sandboxing to Cosmic Files. While I believe thumbnailer sandboxing should be added (as this would help to protect users who like thumbnails), I believe an option to completely disable thumbnailing would be more effective, as sandboxing escapes are always a possibility.

8jef:
I'm also for **disabling all thumbnail image generation** in Cosmic Files. Beside mentioned security concerns, I'd add it takes valuable time and data resources and CPU/GPU power every time one connect a drive with a substantial volume of media files. Moreover, this function actually opens each file to pull a screenshot (and what-not?), and a bit of sound can be heard for each file it scans, every time it scans. Come on people!

Rant

In general, every function involving **automatic anything done on files, apps and software** (all the same to me) **should be OPT-IN ONLY in all scenarios**. We already have an OS that does these things to user pcs, it's called W!#¤0Z3. As a rule of thumb (including thumbnail), Pop/Cosmic, which I use and dearly love but tolerate up to some extent [0] (as a user, not yet a customer) should take actual W!#¤0Z3 play-book, and go (as far possible) the opposite way in such auto behaviour/telemetry/obstruction/obfuscation (except for right-click behaviours, which feels like an afterthought mostly everywhere in Linux Land).

/Rant

You are, as a Team, on the right track, getting there one obstacle at a time. Lot to love here. I like what you do and wish for the Fairy godmother to grant me the means to purchase one of your badass workstation. That will come, hopefully. Big many very gooey kisses to all your cheeks, plus a (gentle?) pinch on the other cheek and a friendly pat on the back for all you bring to the World™, especially at that time of the year! Happy Holidays to All!!!

[0] everything Cosmic is so slow ... I know, it's Alpha™ going Beta™ ...

Issue #1150, opened 2025-08-16
Cosmic-files blocks thumbnails and gallery on remote filesystems


Quackdoc:
Cosmic-files currently checks if the filesystem is a remote one, and if so doesn't allow the preview. Patching out the checks shown below fixes the issue. 

```
        thumbnail_opt: if remote {
            Some(ItemThumbnail::NotImage)
        } else {
            None
        },
```
I did see https://github.com/pop-os/cosmic-files/pull/958, I was wondering if there should be a user configurable list of filesystems to not load on? By default not loading remote filesystems is quite the sledgehammer solution.

PR #1341, opened 2025-11-01
Add thumbnail mode option

infifox:
## Issues addressed

This PR attempts to address the following issues:

- https://github.com/pop-os/cosmic-files/issues/1150
- https://github.com/pop-os/cosmic-files/issues/1253
- https://github.com/pop-os/cosmic-files/issues/1216

It is also a replacement for this unmerged PR:
- https://github.com/pop-os/cosmic-files/pull/1224

## User interface changes

This PR adds an option for the user to configure how they would like to generate thumbnails:

image

The default is the current behavior, where thumbnails are generated only for local files. The user will also be able to choose to generate thumbnails for remote files, or to generate no thumbnails.

## Technical details

This PR replaces https://github.com/pop-os/cosmic-files/pull/1224 with a setting that controls both the option to always generate thumbnails, as well as the option to never generate them. I think it makes sense to have both of these options combined into a single setting, and sets up possibility of adding more complex options in the future.

In order to make this change, I've modified the existing behaviour which preloads the `thumbnail_opt` property for an `Item` struct with a `Some(ItemThumbnail::NotImage)` value to prevent the thumbnail from being generated. Instead, I now include a new `item_location` property which keeps track of whether the item is local, remote, or in the trash. This allows the decision on whether or not to generate a thumbnail to be made within the Tab, avoiding the need to pass the new `thumbnail_mode` value all the way through the file scanning process.

I've also added a new method `item.allow_generate_thumbnail` to collect all the logic for whether or not a thumbnail should be generated in a single place, so it should be easier to expand this logic in the future.

A new message `tab::Message::ThumbConfig` has been added to the Tab, to enable changes to this setting to be immediately propagated to open tabs.

### DIFF
--- a/i18n/en/cosmic_files.ftl
+++ b/i18n/en/cosmic_files.ftl
@@ -297,6 +297,12 @@ type-to-search = Type to Search
 type-to-search-recursive = Searches the current folder and all subfolders
 type-to-search-enter-path = Enters the path to the directory or file
 
+### Thumbnail mode
+thumbnail-mode = Thumbnails
+thumbnail-mode-local = Show for local files only
+thumbnail-mode-all = Show for local and network files
+thumbnail-mode-never = Never show
+
 # Context menu
 add-to-sidebar = Add to sidebar
 compress = Compress
--- a/src/app.rs
+++ b/src/app.rs
@@ -72,7 +72,7 @@ use crate::{
     clipboard::{ClipboardCopy, ClipboardKind, ClipboardPaste},
     config::{
         AppTheme, Config, DesktopConfig, Favorite, IconSizes, TIME_CONFIG_ID, TabConfig,
-        TimeConfig, TypeToSearch,
+        ThumbnailMode, TimeConfig, TypeToSearch,
     },
     dialog::{Dialog, DialogKind, DialogMessage, DialogResult},
     fl, home_dir,
@@ -404,6 +404,7 @@ pub enum Message {
     SearchClear,
     SearchInput(String),
     SetShowDetails(bool),
+    SetThumbnailMode(ThumbnailMode),
     SetTypeToSearch(TypeToSearch),
     SystemThemeModeChange,
     Size(window::Id, Size),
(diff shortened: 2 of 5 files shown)


jackpot51 (approved):

jacobgkau (changes requested):
When I set this to `Show for local and network files`, it makes folders on an SFTP mount load slightly slower, but it's not showing thumbnails for images in the SFTP mount (e.g. screenshots of the same format that have thumbnails shown locally). Watching the network activity in System Monitor, it doesn't look like it's receiving enough data to really be trying to thumbnail everything in the SFTP mount's folders when it's enabled.

Also, I'm wondering if the "always show" and "never show" should be on opposite ends, with "show on local only" between them, since that would be the logical spectrum. The only reason to have "show on local only" on one end is because it's the default. That's a @pop-os/ux question.

jacobgkau:
@maria-komarova Would "file preview" not be ambiguous with the spacebar gallery previewer?

---

I did notice that if I have this set to `Show for local and network files`, then I change it to `Show for local files only`, thumbnails that were already generated over the network still show up. This doesn't match the behavor for local directories; if I change from `Show for local files only` to `Never show`, then thumbnails in a local folder disappear immediately.

I'm observing thumbnails not working on the network mount again after turning them to `Show for local and network files` again-- I've tried navigating out of the SFTP mount and back in, unmounting and re-mounting the SFTP mount, and completely killing/quitting and re-launching COSMIC Files, and completely rebooting, and none of that gets thumbnails working on the SFTP mount immediately.

Levo75:
Can we get thumbnails in the trash bin as well?

KodeBarista:
The feature implementation seems to be working great but I found a couple of bugs which were fixed, as if this branch is based on older codebase.
- Doesn't load system theme (even with v1 theme import & no frost applied)
- `Ctrl+Tab` to switch between tabs doesn't properly work
- The above operation also cause icons to zoom in & out during scroll
- Uses mono font for dialog titles & outdated message
- Sidebar entries become bold when right-clicked

image

PR #1370, opened 2025-11-12
add option to copy path when pressing down shift

FreddyFunk:
https://github.com/user-attachments/assets/9b3d3f72-bf5e-4c38-8778-1e710ea2acfc

### DIFF
--- a/i18n/en/cosmic_files.ftl
+++ b/i18n/en/cosmic_files.ftl
@@ -354,6 +354,7 @@ quit = Quit
 edit = Edit
 cut = Cut
 copy = Copy
+copy-path = Copy path
 paste = Paste
 select-all = Select all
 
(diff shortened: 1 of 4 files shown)


FreddyFunk:
Oh, that's cool! I didn't realize that this was possible. Also, I didn't know that shift + ctrl + c was implemented to copy the file path and only noticed when implementing this feature. Well, then the UX problem I am trying to solve is to let users know that this is possible and how (with which key combination) by making it visible in the UI.

jacobgkau:
@FreddyFunk Maria has given UX approval for this feature on the Shift key here: https://github.com/pop-os/cosmic-files/pull/1564#issuecomment-3880651493 

The code changes look extremely similar. Both PRs have conflicts that need to be resolved. Would you like to rebase this one, since it's already got the Shift functionality?

PR #1378, opened 2025-11-16
Fix windows support

Quackdoc:
Currently windows compilation is broken, the current state of this is just enough to actually get cosmic-files to load on windows again. I plan on cleaning up the added crap and mangled commits later.

- [ ] rustembed which is used for localization breaks when cross compiling, the compile works but the resulting binary is unusable. Not a major issue.
- [ ] xdg-mime is patched to allow compilation on windows, There are a few routes forwards here but I think the best one is to use an alternative mime provider on windows or try to add support for embedded mime. XDG mime on windows is possible, and an installer might be able to make this work. Without mime support 90% of stuff is unusable so this is a major issue.
- [ ] default features are obviously not compatible with windows.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "rust-analyzer.cargo.target": "x86_64-pc-windows-gnu"
+}


Cheong-Lau:
As for `xdg_mime`, the icons you could probably get using the Windows API, and the Mime type using `mime_guess`. But that's a lot of code you'd need to replace

Issue #1215, opened 2025-09-27
Often unable to type into the file browser path on external drives


RoyalOughtness:
Can't type in this: 

<img width="722" height="132" alt="Image" src="https://github.com/user-attachments/assets/bf1454a7-f0a4-4f9b-b04f-30fada4e9952" />

Issue #1099, opened 2025-07-26
[Bug] Hidden files are visible in external drives


KodeBarista:
**Issue/Description:**
When browsing external drives using COSMIC Files, hidden files (those starting with a dot, e.g., `.Trash-1000`) are displayed by default. This behavior is inconsistent with user expectations and differs from how internal volumes or the home directory are treated, where hidden files remain hidden unless explicitly toggled.

**Version:**
```
cosmic-files:
  Installed: 0.1.0~1753472857~24.04~6d9e6c1
```

**Steps to Reproduce:**
- Plug in any external drive (e.g., USB stick or external HDD).
- Open the drive in COSMIC Files.
- Observe that hidden files are visible without toggling the "Show Hidden Files" option.

**Expected Behavior:**
Hidden files on external drives should remain hidden unless the user has explicitly enabled the "Show Hidden Files" option globally or per session.

bobcat141:
Also happens with external partitions from the same SSD, or at least NTFS ones. `.hidden` files also don't work in those partitions.

Here's a video evidence showing the issue on COSMIC Files and comparing it with Thunar:

https://github.com/user-attachments/assets/10f2cbc6-216e-44be-bfce-83494fae1849

iamkartiknayak:
Opening the Data (1TB HDD in the laptop that needs to be mounted) has following issues  (Image 1)

- Doesn't respect show hidden files flag
- No thumbnailing [#1150](#1150 )
- No proper path shown (just the opened dir name) [#1142](#1142)
- Unable to type in the file browser path. [#1215](#1215)
- Missing size for files in the sidebar [#1161](#1161)

If one of those folder in Image 1 is opened with `Open in new tab` option it fixes every problem that's listed in bullet points

<img width="1109" height="738" alt="Image" src="https://github.com/user-attachments/assets/9e4d0629-c045-4680-8f35-8f34297ce53b" />

<img width="1109" height="738" alt="Image" src="https://github.com/user-attachments/assets/f2a8c0c8-9592-4021-90ea-3d7e3dfd1bf3" />

bobcat141:
> If one of those folder in Image 1 is opened with `Open in new tab` option it fixes every problem that's listed in bullet points

Same on my end. Right-clicking a folder and selecting "Open in a new tab" fixes all the issues only for that specific tab. Other tabs opened with `ctrl-t` still display the same problem, however.

PR #1379, opened 2025-11-17
Show metadata, thumbnails and gallery view on mounted drives

FreddyFunk:
https://github.com/pop-os/cosmic-files/pull/958 rightfully introduced a reduction of features for "remote" locations, but "remote" locations were not defined as network locations (smb, cifs, nfs, and so on) but instead defined as everything that is mounted, which would also include USB drives, separate internal drives. This pull request changes:

- metadata, thumbnails, gallery view are now working on mounted locations with direct file system access, but are still disabled for remote locations over the network like network shares for performance reasons
- drives are now mounted differently from the sidebar and use now "/run/media/user/" instead of "file:///media/user/data"
- Clicking on a drive that is not mounted yet in the sidebar, mounts it AND opens it immediately instead of requiring two clicks for that
- improved detection for "remote" file systems

addresses comments of @iamkartiknayak here https://github.com/pop-os/cosmic-files/issues/1099#issuecomment-3525971595
relates to https://github.com/pop-os/cosmic-files/issues/1150
fixes https://github.com/pop-os/cosmic-files/issues/1215

### DIFF
--- a/src/mounter/gvfs.rs
+++ b/src/mounter/gvfs.rs
@@ -40,9 +40,7 @@ fn items(monitor: &gio::VolumeMonitor, sizes: IconSizes) -> MounterItems {
                     gio::Cancellable::NONE,
                 )
                 .ok()
-                .and_then(|info| {
-                    Some(info.boolean(gio::FILE_ATTRIBUTE_FILESYSTEM_REMOTE))
-                })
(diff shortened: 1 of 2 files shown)


leviport (approved):
Great improvements! Thanks!

Issue #1231, opened 2025-10-01
Can't open the context menu in cosmic-files-applet


therealmate:
To reproduce:

1. Open an application that uses cosmic-files-applet (e.g. Firefox download chooser)
2. Right-click, nothing happens instead of the context menu appearing

This may also cause a crash if an app tries to open a popup after doing the above (e.g. cosmic-files right-click, clicking on the time applet, etc) with ```wl_surface#83: error 2: xdg_popup was not created on the topmost popup``` and ```Protocol error 2 on object wl_surface@83:```, but that seems like a cosmic-comp issue)

jacobgkau:
More detailed reproduction steps:

1. Open Firefox and do a right-click -> Save ___ as...
2. Attempt to right-click something within the save dialog (it doesn't work).
3. Close the save dialog and close Firefox. After this is when I see the panel crash when trying to open the battery applet.

PR #1381, opened 2025-11-18
Contains surface

wash2:
part of a fix for https://github.com/pop-os/cosmic-files/issues/1231


jacobgkau (approved):
This gets right-click working in e.g. save dialogs, and stops the panel crash from happening when clicking an applet afterwards.

Skipping the normal regression testing checklist because this is only touching `dialog.rs`.

jackpot51 (approved):

Issue #1340, opened 2025-10-31
Can't enter path in network mounts


jacobgkau:
Tested on version: 4d04ef7

If I connect to an `sftp://` network mount, then navigate into it, attempting to edit the path (including typing at all, if I have typing set to enter the path bar instead of searching) results in the cursor blinking but no additional text being entered into the path bar (which starts with `/run/user/1000/gvfs/sftp:host=192.168...`).

Especially because network mounts are slower to load, it's useful to be able to navigate directly into a specific path. Other file managers support this (Dolphin even has auto-completion of file paths while typing into its path bar in an SFTP mount).

WatchMkr:
Did this PR fix the issue? https://github.com/pop-os/cosmic-files/pull/1379

Cheong-Lau:
@WatchMkr no, but I opened #1390, would appreciate more testing and improvements

PR #1390, opened 2025-11-21
feat(tab): allow entering of network uri as path

Cheong-Lau:
A super hacky way to implement #1340.

Some testing on my Samba server:

https://github.com/user-attachments/assets/fba9351b-5f13-49f5-a3e0-a18dfc996d47

the tab name is properly updated, and history is supported, but tab completion is not (see https://github.com/pop-os/cosmic-files/pull/1390#issuecomment-3564201531).

Closes: #1340

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -3004,7 +3004,12 @@ impl Application for App {
                                             .as_ref()
                                             .map_or_else(|| &tab.location, |x| &x.location);
                                         // Try to add text to end of location
-                                        if let Some(path) = location.path_opt() {
+                                        if let Location::Network(uri, ..) = location {
+                                            let mut uri_string = uri.clone();
+                                            uri_string.push_str(&text);
+                                            tab.edit_location =
+                                                Some(location.with_uri(uri_string).into());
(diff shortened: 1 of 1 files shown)


Cheong-Lau:
@jackpot51 makes sense that actually makes things much simpler. I've pushed a new version

jackpot51 (dismissed):

jacobgkau:
> Sorry, I don't understand what you mean, was this something I need to address or were you just confirming what this PR was doing?

That was just confirming the PR works for doing what it's supposed to.

I'll take another look at this today.

jacobgkau (approved):
This is working as expected now. Thank you for implementing this!

<details>

### Basic navigation

- [ ] Hover over the top item in the folder, then scroll down until it's out of view (while still hovered). On scrolling back up (with the mouse in a different position), the item should not have the hover highlight.
    - Not a regression.
    - Not a regression.

### File operations

- [X] Files can be renamed.

- [X] Sorting options work.

### External filesystems

### Integrations

</details>

PR #1409, opened 2025-12-06
Empty trash as much as possible

### LINKED ISSUE #1134: [Bug] Undeletable file in trash makes emptying trash fail silently
If there is a file in the trash folder (~./local/share/Trash) that the user cannot delete, for example due to a permission issue, then emptying the trash by right-clicking the trash folder and selecting the respective option will silently fail upon encountering that file. No further files will be deleted and the user is not informed that a problem has occured.

I would expect that at least all files except for that one would be deleted, but deletions will stop upon encountering that one undeletable file. The user should also be informed that at least one file could not be deleted. Until now my only interactions with the trash were right click -> empty, so I never realised that it kept filling up.

74togo:
Fixes #1134

### DIFF
--- a/src/operation/mod.rs
+++ b/src/operation/mod.rs
@@ -866,6 +866,8 @@ impl Operation {
                         let items = trash::os_limited::list()
                             .map_err(|e| OperationError::from_err(e, &controller))?;
                         let count = items.len();
+                        let mut errors: Vec<trash::Error> = Vec::new();
+
                         for (i, item) in items.into_iter().enumerate() {
                             futures::executor::block_on(async {
                                 controller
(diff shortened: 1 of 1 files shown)


jacobgkau (dismissed):
I was not able to replicate https://github.com/pop-os/cosmic-files/issues/1134 on the most recent version of COSMIC Files without this change; if I trash a file, then manually go into the Trash directory and set the file's owner to root, emptying the trash in COSMIC Files still deleted the file successfully.

However, emptying the trash still works after this change, and regression testing has passed.

<details>
<summary>Checklist:</summary>

## Checklist

### Basic navigation

- [ ] Hover over the top item in the folder, then scroll down until it's out of view (while still hovered). On scrolling back up (with the mouse in a different position), the item should not have the hover highlight.
    - Not a regression.
    - Not a regression.

### File operations

- [ ] Navigate to the parent folder, create another empty file, then drag it into the folder.
    - Not a regression.
- [X] Files can be renamed.

- [X] Sorting options work.

### External filesystems

### Integrations

</details>

Issue #1408, opened 2025-12-05
Inhibit screensaver/sleep during file operations


nathansgithub:
It looks like there is already a TODO for this, but I thought it was important to call out. I've lost some data because I was transferring files to an external drive when the system went to sleep.

https://github.com/pop-os/cosmic-files/blob/6fa6d305e054b99dabd89b4c0f14ff520897e58b/src/app.rs#L6129

joshuamegnauth54:
Thanks for raising this issue. I recently implemented this exact feature for COSMIC player in pop-os/cosmic-player#185. I'll attempt to port it over this weekend.

nathansgithub:
@joshuamegnauth54 I saw your PR! Is there anything I can help with?

joshuamegnauth54:
My main problem is testing the PR. 😅 I wrote a similar, working patch for COSMIC Player a while ago. This patch *should* work but I want to be sure before I undraft it. Also, I probably have to rebase at this point. 😆 

I wrote a small FUSE filesystem that pauses during file operations to help with testing, but it seemed to block COSMIC Files' executor. Interestingly, I just saw that there's a PR that seems to fix that exact issue, so maybe I can revive my PR soon.

* #1879

Someone else seemed to merge my PR into their fork, so maybe it actually works: https://github.com/Lcstyle/cosmic-files/commit/419df460ddc12dfb28020fd6ac5a882e736e3209

nathansgithub:
@joshuamegnauth54 I was wondering if testing would be difficult. Maybe an easier test would be just copying a large file to a slow USB drive? I haven't tried it yet, but you should be able to change the timeout to something super short by editing the file at `$HOME/.config/cosmic/com.system76.CosmicIdle/v1/screen_off_time`

PR #1410, opened 2025-12-06
Prevent suspension during file operations

joshuamegnauth54:
Closes: #1408
Partially a port of: pop-os/cosmic-player#185

---

WIP and placeholder.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1459,6 +1459,7 @@ name = "cosmic-files"
 version = "0.1.0"
 dependencies = [
  "anyhow",
+ "ashpd 0.12.0",
  "bzip2",
  "chrono",
  "compio",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,7 @@ rust-version = "1.85"
 
 [dependencies]
 anyhow = "1"
+ashpd = { version = "0.12", optional = true }
 chrono = { version = "0.4", features = ["unstable-locales"] }
 icu = { version = "2.1.1", features = ["compiled_data"] }
 cctk = { git = "https://github.com/pop-os/cosmic-protocols", package = "cosmic-client-toolkit", rev = "d0e95be", optional = true }
@@ -103,6 +104,7 @@ default = [
     "notify",
     "wgpu",
     "wayland",
+    "xdg-portal",
 ]
 dbus-config = ["libcosmic/dbus-config"]
 desktop = ["libcosmic/desktop", "dep:cosmic-mime-apps", "dep:xdg"]
@@ -114,6 +116,7 @@ jemalloc = ["dep:tikv-jemallocator"]
 notify = ["dep:notify-rust"]
 wayland = ["libcosmic/wayland", "dep:cctk", "dep:wayland-client"]
 wgpu = ["libcosmic/wgpu"]
+xdg-portal = ["dep:ashpd"]
 
 [profile.dev]
 opt-level = 1
--- a/src/app.rs
+++ b/src/app.rs
@@ -731,6 +731,8 @@ pub struct App {
     tab_drag_id: DragId,
     auto_scroll_speed: Option<i16>,
     file_dialog_opt: Option<Dialog<Message>>,
+    #[cfg(feature = "xdg-portal")]
+    pub inhibit: tokio::sync::watch::Sender<bool>,
 }
 
 impl App {
@@ -2072,6 +2074,20 @@ impl App {
 
         false
     }
+
+    /// Allow user to suspend or log out.
+    ///
+    /// Basically, undo [`Self::inhibit_suspension`].
+    fn allow_suspend(&self) {
+        #[cfg(feature = "xdg-portal")]
+        let _ = self.inhibit.send(false);
+    }
+
+    /// Prevent suspending during active file operations.
+    fn inhibit_suspend(&self) {
+        #[cfg(feature = "xdg-portal")]
+        let _ = self.inhibit.send(true);
+    }
 }
 
 /// Implement [`Application`] to integrate with COSMIC.
@@ -2153,6 +2169,15 @@ impl Application for App {
                 ),
             ]);
 
+        // The inhibit task must be created here since we need a runtime and libcosmic spawns
+        // the runtime - not main!
+        #[cfg(feature = "xdg-portal")]
+        let inhibit = {
+            let (tx, rx) = tokio::sync::watch::channel(false);
+            std::mem::drop(tokio::spawn(crate::xdg_portals::inhibit(rx)));
+            tx
+        };
+
         let mut app = Self {
             core,
             about,
@@ -2204,6 +2229,8 @@ impl Application for App {
             file_dialog_opt: None,
             #[cfg(all(feature = "wayland", feature = "desktop-applet"))]
             layer_sizes: FxHashMap::default(),
+            #[cfg(feature = "xdg-portal")]
+            inhibit,
         };
 
         let mut commands = vec![app.update_config()];
@@ -6126,7 +6153,7 @@ impl Application for App {
         }));
 
         if !self.pending_operations.is_empty() {
-            //TODO: inhibit suspend/shutdown?
+            self.inhibit_suspend();
 
             if self.core.main_window_id().is_some() {
                 // Force refresh the UI every 100ms while an operation is active.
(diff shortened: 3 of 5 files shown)


joshuamegnauth54:
This should work but I'm having trouble figuring out how to test it. I wrote a small FUSE file system that delays writes which will hopefully be a useful way to test it.

```rust
// SPDX-License-Identifier: MIT

// USAGE:
// cargo add fuser
// cargo add libc
// cargo run -- /empty/folder delay-in-seconds
//
// The file system only supports writing files to the root mount point. Writes are delayed for
// the seconds passed in above.

use std::{
    collections::BTreeMap,
    ffi::OsStr,
    io::{self, Cursor, Seek, SeekFrom, Write},
    path::Path,
    sync::LazyLock,
    thread,
    time::{Duration, SystemTime},
};

use fuser::{
    FileAttr, FileType, Filesystem, MountOption, ReplyAttr, ReplyCreate, ReplyData, ReplyDirectory,
    ReplyEntry, ReplyWrite, Request,
};
use libc::{EINVAL, ENOENT};

const BLOCK_SIZE: u64 = 4096;
static ROOT_DOT: LazyLock<SlowNode> = LazyLock::new(|| SlowNode {
    attr: attr(1, 0, FileType::Directory, 0o755),
    data: Vec::new(),
    name: ".".into(),
});
static ROOT_DOTDOT: LazyLock<SlowNode> = LazyLock::new(|| SlowNode {
    attr: attr(1, 0, FileType::Directory, 0o755),
    data: Vec::new(),
    name: "..".into(),
});
const TTL: Duration = Duration::from_secs(1);

struct SlowFS {
    delay: Duration,
    entries: BTreeMap<u64, SlowNode>,
    file_handle: u64,
}

impl SlowFS {
    pub const fn new(delay: Duration) -> Self {
        Self {
            delay,
            entries: BTreeMap::new(),
            file_handle: 2,
        }
    }
}

impl Filesystem for SlowFS {
    fn getattr(&mut self, _req: &Request<'_>, ino: u64, _: Option<u64>, reply: ReplyAttr) {
        if ino == 1 {
            reply.attr(&TTL, &ROOT_DOT.attr)
        } else if let Some(node) = self.entries.get(&ino) {
            reply.attr(&TTL, &node.attr)
        } else {
            reply.error(ENOENT)
        }
    }

    fn create(
        &mut self,
        _req: &Request<'_>,
        parent: u64,
        name: &OsStr,
        mode: u32,
        _umask: u32,
        _flags: i32,
        reply: ReplyCreate,
    ) {
        if parent != 1 {
            reply.error(ENOENT);
            return;
        }

        let ino: u64 = self.entries.keys().max().copied().unwrap_or_else(|| {
            assert!(self.entries.is_empty());
            2
        });

        let attr = attr(ino, 0, FileType::RegularFile, mode as u16);
        self.entries.insert(
            ino,
            SlowNode {
                attr,
                data: Vec::new(),
                name: name.to_string_lossy().to_string(),
            },
        );

        // Pretty sure this will never overflow in a test file system!
        self.file_handle += 1;

        reply.created(&Duration::new(0, 0), &attr, 0, self.file_handle, 0);
    }

    fn lookup(&mut self, _req: &Request<'_>, parent: u64, name: &OsStr, reply: ReplyEntry) {
        let name = name.to_string_lossy();
        println!("[lookup] {name}");

        // Only one root directory supported.
        if parent != 1 {
            reply.error(ENOENT);
            return;
        }

        if let Some(node) = self.entries.values().find(|node| node.name == name) {
            reply.entry(&Duration::new(0, 0), &node.attr, 0);
        } else {
            reply.error(ENOENT)
        }
    }

    fn readdir(
        &mut self,
        _req: &Request<'_>,
        ino: u64,
        _fh: u64,
        offset: i64,
        mut reply: ReplyDirectory,
    ) {
        println!("[readdir] inode: {ino}");

        // For simplicity, only support a flat directory structure.
        if ino != 1 {
            reply.error(ENOENT);
            return;
        }

        let default = [&*ROOT_DOT, &*ROOT_DOTDOT];
        for (i, node) in default
            .into_iter()
            .chain(self.entries.values())
            .skip(offset as usize)
            .enumerate()
        {
            let full = reply.add(
                node.attr.ino,
                offset + i as i64 + 1,
                node.attr.kind,
                &node.name,
            );

            if full {
                break;
            }
        }

        reply.ok();
    }

    fn write(
        &mut self,
        _req: &Request<'_>,
        ino: u64,
        _fh: u64,
        offset: i64,
        data: &[u8],
        _write_flags: u32,
        _flags: i32,
        _lock_owner: Option<u64>,
        reply: ReplyWrite,
    ) {
        println!("[write] inode: {ino}");

        if offset < 0 {
            eprintln!("[write] Negative offset");
            reply.error(EINVAL);
            return;
        }

        let Some(node) = self.entries.get_mut(&ino) else {
            eprintln!("[write] inode doesn't exist: {ino}");
            reply.error(ENOENT);
            return;
        };

        println!(
            "[write] Writing {} bytes to inode {} at offset {} with {} secs of delay",
            data.len(),
            ino,
            offset,
            self.delay.as_secs()
        );

        let mut file = Cursor::new(&mut node.data);
        if let Err(e) = file.seek(SeekFrom::Start(offset as u64)) {
            eprintln!("[write] Error seeking to inode {ino} at {offset}: {e}");
            reply.error(EINVAL);
            return;
        }

        thread::sleep(self.delay);

        if let Err(e) = file.write_all(data) {
            eprintln!("[write] Error writing to inode {ino}: {e}");
            reply.error(EINVAL);
            // I'm not cleaning up after the error. It's a temp file system after all!
            return;
        }

        node.attr.size = node.data.len().try_into().expect(
            "What kind of fancy machine are you using that a usize doesn't fit into a u64?",
        );
        node.attr.blocks = node.attr.size.div_ceil(BLOCK_SIZE);
        // The buffer size type is technically a u32 so data will never be larger anyway.
        reply.written(data.len() as _);
    }

    fn read(
        &mut self,
        _req: &Request<'_>,
        ino: u64,
        _fh: u64,
        offset: i64,
        size: u32,
        _flags: i32,
        _lock_owner: Option<u64>,
        reply: ReplyData,
    ) {
        println!("[read] inode: {ino} offset: {offset} size: {size}");

        let Some(node) = self.entries.get(&ino) else {
            eprintln!("[read] inode not found: {ino}");
            reply.error(ENOENT);
            return;
        };

        let size = size as usize;
        let offset = match offset.try_into() {
            Ok(offset) if offset + size < node.data.len() => offset,
            _ => {
                eprintln!("[read] offset out of bounds");
                reply.error(EINVAL);
                return;
            }
        };

        let Some(chunk) = node.data.get(offset..offset + size) else {
            eprintln!("[read] offset is out of bounds");
            reply.error(EINVAL);
            return;
        };
        reply.data(chunk);
    }
}

struct SlowNode {
    attr: FileAttr,
    data: Vec<u8>,
    name: String,
}

fn attr(ino: u64, size: u64, kind: FileType, perm: u16) -> FileAttr {
    let uid = unsafe { libc::getuid() };
    assert_ne!(0, uid);
    let gid = unsafe { libc::getgid() };
    assert_ne!(0, gid);

    let blocks = size / BLOCK_SIZE;

    FileAttr {
        ino,
        size,
        blocks,
        atime: SystemTime::UNIX_EPOCH,
        mtime: SystemTime::UNIX_EPOCH,
        ctime: SystemTime::UNIX_EPOCH,
        crtime: SystemTime::UNIX_EPOCH,
        kind,
        perm,
        nlink: 1,
        uid,
        gid,
        rdev: 0,
        blksize: 4096,
        flags: libc::O_RDWR as _,
    }
}

fn main() -> io::Result<()> {
    let (fs, mount_point) = {
        let mut args = std::env::args().skip(1);
        let mount_point = args
            .next()
            .map(|mount| Path::new(&mount).canonicalize().unwrap())
            .expect("First argument should be the mount point");
        let fs = args
            .next()
            .and_then(|s| s.parse().ok())
            .map(Duration::from_secs)
            .map(SlowFS::new)
            .expect("Second argument should be the write delay in seconds");

        (fs, mount_point)
    };

    println!("Mounting at {}", mount_point.to_string_lossy());

    fuser::mount2(
        fs,
        mount_point,
        &[
            MountOption::AutoUnmount,
            MountOption::FSName("slowfs".into()),
            MountOption::NoAtime,
            MountOption::NoDev,
            MountOption::NoExec,
            MountOption::NoSuid,
            MountOption::RW,
        ],
    )
}
```

PR #1450, opened 2025-12-17
fix: support absolute paths in .desktop file Icon field

nilltadios:
## Summary

Per the [FreeDesktop Desktop Entry spec](https://specifications.freedesktop.org/desktop-entry-spec/latest/recognized-keys.html), the `Icon` field can be either:
- An icon name (looked up in the current theme)
- An absolute path to an image file

Previously, all `Icon` values were passed to `widget::icon::from_name()`, which only works for icon names. This caused `.desktop` files with absolute icon paths to display placeholder icons instead of the actual image.

## Changes

- Added `desktop_icon_handle()` helper function that checks if the icon path is absolute and exists
- Uses `widget::icon::from_path()` for absolute paths
- Uses `widget::icon::from_name()` for icon names (existing behavior)
- Updated both `item_from_gvfs_info()` and `item_from_entry()` to use the helper

## Test plan

1. Create a `.desktop` file with an absolute icon path:
   ```ini
   [Desktop Entry]
   Name=Test
   Type=Application
   Exec=true
   Icon=/path/to/image.png
   ```
2. Place on desktop or view in file manager
3. Verify the custom icon displays instead of a placeholder

Fixes: https://github.com/pop-os/cosmic-epoch/issues/2697

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -5589,7 +5589,7 @@ impl Application for App {
         }
         let finished = count - running;
         total_progress /= count as f32;
-        if running > 1 || finished > 0 {
+        if running >= 1 && (running > 1 || finished > 0) {
             if finished > 0 {
                 title = fl!(
                     "operations-running-finished",


git-f0x:
I'm curious when this regressed, since it used to work fine. Maybe [this](https://github.com/pop-os/libcosmic/commit/639326fcc31a95a0c7a9a5bb56f1ed4d53530f26)?

PR #1471, opened 2025-12-27
add type-to-select keyboard option

wowitsjack:
adds a third option in settings for type-to-search behavior - typing letters selects the first matching file instead of opening search. accumulates typed chars and resets after 1 second timeout. works like windows explorer.


maria-komarova:
I'm good with it as is if there is no easy and fast way to solve the scroll issue.

PR #1483, opened 2025-12-30
fix: scroll to selected file when opening file location

74togo:
When opening the location of a file, the file manager failed to scroll to
the target file because item.rect_opt was only populated during
rendering, which hadn't occurred yet.

Changes:
- Add item_rect_grid/list() functions to calculate item positions without
  requiring prior rendering
- Modify select_focus_scroll() to fall back to item_rect_grid/list() when
  rect_opt is unavailable
- Add should_scroll parameter to update_tab/rescan_tab to trigger
  scrolling only when opening a new location (not during rescans or
  other tab updates)

The scroll_to task is now batched with the clipboard read task in the
TabRescan message handler when scrolling is requested.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -424,6 +424,7 @@ pub enum Message {
         Option<tab::Item>,
         Vec<tab::Item>,
         Option<Vec<PathBuf>>,
+        bool,
     ),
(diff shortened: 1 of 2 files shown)


jackpot51:
This was already done in #1482

Issue #1470, opened 2025-12-26
Thumbnailer infinitely creates new thumbnails while in the ~/.cache/thumbnails


ashxx64:
## Issue

While in the subdirectories of this directory, COSMIC will see the properly created thumbnails, such as for pictures in ~/Pictures. However, COSMIC will then create thumbnails of the thumbnails. Then it will create a thumbnail of the thumbnail of the thumbnail and so on.

## Severity

The issue is not terrible as COSMIC does not create thumbnails of the entire directory at once, only a subset, the visible things. A lot of new thumbnails will be created if you scroll, but will settle down once you stop scrolling.

## Possible Solution

GNOME addresses this issue by not thumbnailing any directories that have "thumbnails" in the path, such as ~/.cache/thumbnails/x-large or ~/Pictures/custom-dir/thumbnails/blah.

## System Details
OS: Fedora Linux 43.20251225.0 (COSMIC Atomic) x86_64
Files: cosmic-files-1.0.0-1.fc43.x86_64

jacobgkau:
@ashxx64 Are you still experiencing this issue? I'm not able to recreate it. Someone opened a PR trying to fix it, but before the PR was finished, they also said they were no longer able to recreate the issue on the latest version.

PR #1490, opened 2026-01-01
Don't thumbnail thumbnails

joshuamegnauth54:
Closes: #1470

According to the issue, GNOME just skips files under any folder named "thumbnails." ~~I implemented the same logic here.~~ This implementation only skips the thumbnailing folder as defined by the [Thumbnail Managing Standard](https://specifications.freedesktop.org/thumbnail/latest/) of Freedesktop.

---

I tested this by browsing to folders with thumbnails in my cache and observing that thumbnails weren't created.

### DIFF
--- a/src/tab.rs
+++ b/src/tab.rs
@@ -91,7 +91,7 @@ use crate::{
     mounter::MOUNTERS,
     mouse_area,
     operation::{Controller, OperationError},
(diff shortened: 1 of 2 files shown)


jackpot51:
This should check if inside the thumbnails directory, in my opinion, instead of if the path contains thumbnails

PR #1498, opened 2026-01-04
fix: Update data on preview in desktop mode

jpttrssn:
Updates data including directory sizes during preview in desktop mode. Re-calculates directory sizes each time a new preview window is created.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -3669,7 +3669,10 @@ impl Application for App {
                                 id,
                                 Window::new(WindowKind::Preview(entity_opt, preview_kind)),
                             );
-                            return command.map(|_id| cosmic::action::none());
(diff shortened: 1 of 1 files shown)


jacobgkau (changes requested):
This fixes the bug in the preview dialog on the desktop, but it seems to break thumbnail display in any folder in the regular COSMIC Files app. They work fine when going back to master on cebb8f2c66bae0fcda80740f1807be258fa63f90, but don't work on this branch on 53bd6f7. As the only two commits in master that aren't on this branch were very self-contained Italian translation updates, it seems like that makes it a regression in this branch.

jpttrssn:
Good catch. Fixed.

PR #1499, opened 2026-01-04
feat: Add open button to multi preview

jpttrssn:
Will open all currently selected items. Directories will open in separate tabs for both app and desktop mode when multiple items are selected.

### DIFF
--- a/src/tab.rs
+++ b/src/tab.rs
@@ -3299,7 +3299,13 @@ impl Tab {
                                     .take(max_real - min_real + 1)
                                 {
                                     if let Some(item) = items.get_mut(index) {
-                                        item.selected = true;
+                                        if item.hidden {
+                                            if self.config.show_hidden {
+                                                item.selected = true;
+                                            }
+                                        } else {
+                                            item.selected = true;
+                                        }
                                     }
                                 }
                             }


jackpot51 (approved):

jacobgkau (changes requested):
This works for multiple files.

The behavior with one folder and multiple files seems to be as expected.

When clicking this with multiple folders, it looks like only the last folder is opened. Can you either make the folders all open in separate tabs, or hide the button if multiple folders are selected?

jacobgkau (changes requested):
This has sort of exposed a bug where hidden files and directories between non-hidden ones are selected when doing e.g. shift-click between items, even if hidden files/directories aren't set to show. That bug itself is not a regression, but it only presented with files and not directories before, since only one directory would ever open when pressing Enter on them before, whereas directories will open in separate tabs now. I've opened an issue to track that: https://github.com/pop-os/cosmic-files/issues/1502 We don't need to block this PR on that issue, though.

Aside from that quirk, I'm afraid 043b481 breaks the behavior when pressing Enter with only a single directory selected-- even if only a single directory is selected, pressing Enter now opens it in a background tab. We still need to open a selected single directory in the same tab, since this is a common navigation pattern.

jackpot51 (approved):

PR #1501, opened 2026-01-06
fix: culled items not appearning after window resize

hojjatabdollahi:
When a window resizes larger such that content no longer needs scrolling, scroll_opt retains its stale offset value. Since no scroll event fires, visible_rect was calculated with an out-of-bounds offset, causing items at the top to be culled and not rendered.

Steps to recreate:
1- open a directory with many items but not enough to scroll
2- Press meta+t to open another window next to it which halves the size of this window therefore it now needs to scroll.
3- scroll down a bit and select a file at the bottom
4- now close the other window, so cosmic-files would go back to its original size which didn't need scrolling
5- The items that you scrolled down are now not rendered anymore

https://github.com/user-attachments/assets/bcceeb4c-90df-452a-aa3f-3161457dd192

Fix by caching content_height from each render pass and using it to clamp scroll_opt when building visible_rect.

### DIFF
--- a/src/tab.rs
+++ b/src/tab.rs
@@ -2569,6 +2569,7 @@ pub struct Tab {
     pub mode: Mode,
     pub scroll_opt: Option<AbsoluteOffset>,
     pub size_opt: Cell<Option<Size>>,
+    pub content_height_opt: Cell<Option<f32>>,
     pub viewport_opt: Option<Rectangle>,
     pub item_view_size_opt: Cell<Option<Size>>,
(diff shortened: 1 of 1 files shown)


jacobgkau (approved):
I was able to recreate the issue this is fixing once per launch in grid view. I can confirm this fixes the issue.

<details>

## Checklist

### Basic navigation

    - Not a regression.

### File operations

- [X] Files can be renamed.

- [X] Sorting options work.

### External filesystems

### Integrations

</details>

PR #1503, opened 2026-01-07
Path editing improvements

jackpot51:
Implemented completion using Tab, nested path completion, and proper handling of `..`

### DIFF
--- a/src/tab.rs
+++ b/src/tab.rs
@@ -1469,8 +1469,10 @@ impl Location {
             if let Ok(canonical) = fs::canonicalize(&path) {
                 path = canonical;
             }
-            // Add trailing slash if location is a path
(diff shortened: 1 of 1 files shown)


wash2 (approved):

jacobgkau (commented):
The last commit to automatically move onto the next level if there's only one possible auto-completion is working well.

When tab-completing file names, they also get a trailing slash. We'd probably want to only do that for directory names.

I'm also noticing that hidden files & directories show up in the autocompletion drop-down even if showing hidden files is disabled. That's not a problem and somewhat makes sense for directories, but it results in not being taken to the file when selecting a hidden file. So we probably want to hide hidden files (at least) from autocompletion if they're not being shown, right?

PR #1520, opened 2026-01-12
Add setting open-with and permissions for multiple selected items

jpttrssn:
### Open-with
Adds the ability to set the open-with setting for multiple selected items. This setting will only appear if all selected items have the same mime type.

### Permissions
This adds a `Message::ShiftPermissions` to handle setting permissions for user, group or other for either 1 or more items and a `Command::SetMultiplePermissions` to set permissions on multiple items at the same time.

The permission dropdown will only have a selection if all selected items have the same permission, otherwise it will be empty but still allow changing.

Up to 5 owners and groups will be displayed for all selected items with an ellipses if there are more.

The `OperationSelection` for setting permissions now also returns the path as selected such that the tab will be re-scanned such that the dropdown will be updated correctly.

### Join operations
Allow for joining operations into a single Task that will produce a single `Message:PendingResults` message such that multiple `Message::PendingComplete` and `Message::PendingError` messages can be handled together to, for example like with `Command::SetMultiplePermissions`, re-scan once or show a single error dialog with multiple errors.

### Regression testing
Since these changes affect single operation handling in general (wrapped in a loop) these should be tested as well.

This is meant to be applied to #1499

### DIFF
--- a/i18n/en/cosmic_files.ftl
+++ b/i18n/en/cosmic_files.ftl
@@ -131,6 +131,7 @@ open-with = Open with
 owner = Owner
 group = Group
 other = Other
+mixed = Mixed
 ### Mode 0
 none = None
 ### Mode 1 (unusual)
--- a/i18n/sv/cosmic_files.ftl
+++ b/i18n/sv/cosmic_files.ftl
@@ -101,6 +101,7 @@ open-with = Öppna med
 owner = Ägare
 group = Grupp
 other = Andra
+mixed = Blandade
 # Listvy
 name = Namn
 modified = Ändrad
--- a/src/tab.rs
+++ b/src/tab.rs
@@ -6445,17 +6445,6 @@ impl Tab {
 
         #[cfg(unix)]
         {
-            // The following is a bit of a hack to make the dropdown selector
-            // fill the available space when selected is None so it can be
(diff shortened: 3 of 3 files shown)


jacobgkau:
I'll take a look at this once https://github.com/pop-os/cosmic-files/pull/1499 has been merged.

jpttrssn:
Rebased.

jackpot51 (approved):

jacobgkau:
Displaying and setting the default application and the user/group/other permissions work, both for individual items and multiple items selected.

When items with different permissions are selected, we currently get a blank label on the dropdown, as you mentioned in the PR description:

Screenshot_2026-01-14_15-01-32
Screenshot_2026-01-14_15-01-38

I'm wondering if having some kind of label, like `(Various)`, would be clearer than just leaving the drop-down blank. @pop-os/ux Do you have thoughts on this?

jpttrssn:
Latest changes update libcosmic, add `mixed` as part of the language file and add the placeholder text for the dropdowns using `mixed`.

jackpot51 (dismissed):

jacobgkau:
The build failed because I'd messed something up while trying to resolve the conflicts, should work now.

PR #1534, opened 2026-01-16
Add file checksums to properties panel

### LINKED ISSUE #399: [Feature request] Ability to optionally see file hashes in properties dialog
I know this can just be done from the terminal, but it's mostly because [Dolphin](https://github.com/KDE/dolphin) allows the user to see the MD5, SHA1, SHA256 and SHA512 checksums of a given file in the properties dialog, and I personally find it extremely convenient.

FreddyFunk:
closes #399

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1425,7 +1425,6 @@ dependencies = [
  "rust-embed",
  "rustc-hash 2.1.2",
  "serde",
- "sha1 0.10.6",
  "sha2 0.10.9",
  "shlex 1.3.0",
  "slotmap",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1425,6 +1425,8 @@ dependencies = [
  "rust-embed",
  "rustc-hash 2.1.2",
  "serde",
+ "sha1 0.10.6",
+ "sha2 0.10.9",
  "shlex 1.3.0",
  "slotmap",
  "tar",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,6 +56,8 @@ slotmap = "1.1.1"
 recently-used-xbel = "1.2.0"
 zip = "8"
 md-5 = "0.10.6"
+sha1 = "0.10"
+sha2 = "0.10"
 png = "0.18"
 jxl-oxide = { version = "0.12.5", features = ["image"] }
 num_cpus = "1.17.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,7 +56,6 @@ slotmap = "1.1.1"
 recently-used-xbel = "1.2.0"
 zip = "8"
 md-5 = "0.10.6"
-sha1 = "0.10"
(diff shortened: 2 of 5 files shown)


edfloreshz:
Looks like this may need to be rebased before merging

FreddyFunk:
https://github.com/user-attachments/assets/53e0e57c-3922-415b-b9ff-6e04cc76aa8c

FreddyFunk:
@edfloreshz rebased to latest master and added:

- support for ellipses in hash previews
- loading animation for checksums
- copy to clipboard for checksums

FreddyFunk:
@maria-komarova what do you think about this?

FreddyFunk:
The example video is done on a very large file to demonstrate the loading animation. Checksum calculation for smaller files feel instant.

edfloreshz:
@FreddyFunk Are the checksums cached, or do they need to be recalculated every time?

edfloreshz:
@FreddyFunk Could the checksums be calculated individually? You might only be interested in a single one, which could reduce calculation time and provide more user control. 

The checksum UI could be always displayed, and each type would have a calculate button. You could even add a calculate all button if desired.

maria-komarova:
We have discussed it inside the team and we think SHA256 might be more useful than others so we'd like to start there. 
We can add it to the List View with permissions:
image

FreddyFunk:
@edfloreshz Right now they're not persisted. The result is held in memory on the item while its properties are open, but it's recalculated whenever the directory is rescanned or you reopen properties.

@edfloreshz Currently I read the file **once** and feed all four hashers in a single pass. Reading the file (disk I/O) is the bottleneck, so computing extras is nearly free and not noticeable. So it does not matter if I calculate just SHA256 or MD5 + SHA1 + SHA256 + SHA512, both cases take the same amount of time since the disk I/O is the bottleneck.

@maria-komarova thanks for the mock up! I implemented it like you suggested. I kept the "one read + many hashers" architecture, but only have SHA256 in calculated for now. So this can be easily expanded.



https://github.com/user-attachments/assets/e3d1ad6d-eec4-4dbc-961a-658462ce2153

maria-komarova:
Looks good.

jklgrasso (approved):
<details>
<summary>Regression testing passed</summary>

## Checklist

### Basic navigation

- [x] Hover over the top item in the folder, then scroll down until it's out of view (while still hovered).
- [x] Right-click an item in the sidebar. No visual change should occur with the rest of the items.

### File operations

- [x] Navigate to the parent folder, create another new file, then drag it into the created folder.
- [x] Files can be renamed.

- [x] Image and video thumbnails generate & display in local folders.
- [x] Sorting options work.

### External filesystems

### Integrations

- [x] Desktop icons display as expected
- [x] Drag-and-drop into Firefox works


</details>

mmstick (approved):

PR #1541, opened 2026-01-19
fix: Prefer magic bytes to determine extraction logic

### LINKED ISSUE #1522: Unpacking tar fails with "failed to iterate over files.
When trying to extract CQue_v4.0.1_Linux_64_EN.tar, downloaded from the Canon Website under [here](https://files.canon-europe.com/files/soft01-48570/Driver/CQue_v4.0.1_Linux_64_EN.tar ) I get the error "Failed Operation" with "failed to iterate over archive":

Image

Unpacking the file in terminal with `tar -xvf CQue_v4.0.1_Linux_64_EN.tar` works without issue.

norepro:
Extracting files uses the filename to determine the MIME type and thus the crates to invoke for extracting the file. This causes an error if the file has an incorrect extension, e.g., `.tar` instead of `.tar.gz`.

Fixing the file extension works, but it is a bad user experience, especially compared to other file managers (Dolphin, Nautilus) and utilities (tar) that succeed in the same scenario.

When extracting a file, try to determine the file type by looking at its magic bytes. If that logic fails for any reason, fallback to the filename logic.

Checking magic bytes is technically slower than simple path logic, but this is only invoked on a single file when the user has confirmed they want to extract it anyway. This logic is _not_ used in any other scenario, e.g. determining icons to show.

Fixes #1522


norepro:
Merged with master. It looks like my bug on xdg-mime-rs hasn't had any interaction, so I do not think that we will get the reduced code in this PR any time soon. Still works and fixes the bug, though!

Issue #1268, opened 2025-10-09
[Bug] Open with default program seemingly ignored


mattrunyon:
When I open a zip file via double click or select and enter, it opens using `Files` instead of `File Roller`. `File Roller` is shown as the default in the `Open with...` menu. I want the default `File Roller` because I want to see the files before unzipping

There also doesn't seem to be a way to set the default from the "Open with..." menu

<img width="574" height="688" alt="Image" src="https://github.com/user-attachments/assets/0f5d1f17-fd42-4960-9738-b5cac7053c07" />

WatchMkr:
- COSMIC Files doesn't show up as an option for archives
- Show Details shows "File Roller" same as "Open with..." but the option isn't honored
- Should we have an archive option in Settings > Applications > Default Applications?

Xemnas97:
I can confirm this behavior is still present on the latest updates.

System: Pop!_OS 24.04 (COSMIC Alpha)
File Manager: cosmic-files
Archive Manager: File Roller 44.3

`xdg-mime query default application/zip` correctly returns `org.gnome.FileRoller.desktop`, but double-clicking a zip file inside COSMIC Files ignores this and triggers the internal extraction prompt/handler instead of opening the application.

synthfi:
also seeing this. I can't change my preferred image viewer to be anything  but chromium for whatever reason (maybe because it's the first application under the cosmic settings dropdown?). Using Asahi Fedora Remix 42. using `open` or `xdg-open` in the terminal respects my choice correctly

mmstick:
This should be fixed as of #1857

omnireq:
[#1857](https://github.com/pop-os/cosmic-files/pull/1857) doesn't seem to have fixed it. I just had a ticket with it where double clicking is still doing extract to rather than opening in fileroller (default). I tested on my install and got the same behavior.

jungseung390:
@mmstick Is anyone working on this still? Looks like this still isn't fixed. Cosmic-files is still opening .c files with cosmic text editor instead of my default in latest 1.7.0.

PR #1555, opened 2026-01-22
feat: allow configuring archive app

thegabriele97:
## Change default behavior when opening archive files

fixes: #1268 

Opening an archive should mean *looking inside it first* — extracting should be an explicit action.

# Summary

This PR changes how COSMIC Files opens archive files.

**Current behavior:**  
- Only archives selected → extraction dialog opens automatically  
- Mixed selection (archive + other files) → archive opens in archive manager  

**Problem:**  
- Users often want to inspect contents before extracting, grab a single file, or check structure/safety  
- Inconsistent UX between single-archive and mixed selections  

**Proposed behavior:**  
- Double-click/open → archives always open in archive manager  
- “Extract To…” remains available via right-click  
- Default action becomes safe, predictable, and consistent  
- **If there are no suitable apps, it falls back to the extract_to dialog**

**Benefits:**  
- Non-destructive first action  
- Extraction is explicit  
- Same behavior regardless of selection type  
- Aligns with user expectations and other file managers  
- No functionality removed; extraction still one click away  

**Conclusion:**  
Opening an archive now means browsing its contents first; extraction is an explicit, deliberate action.

edit: summarized

### DIFF
--- a/res/com.system76.CosmicFiles.desktop
+++ b/res/com.system76.CosmicFiles.desktop
@@ -9,16 +9,3 @@ Icon=com.system76.CosmicFiles
 Categories=COSMIC;Utility;FileManager;
 Keywords=Folder;Manager;
 MimeType=inode/directory;application/gzip;application/x-compressed-tar;application/x-tar;application/zip;application/x-bzip;application/x-bzip-compressed-tar;application/x-bzip2;application/x-bzip2-compressed-tar;application/x-xz;application/x-xz-compressed-tar;
-
-Actions=Extract;ExtractTo;
-
-[Desktop Action Extract]
-Name=COSMIC Files Extract
-Exec=cosmic-files --extract %U
-NoDisplay=true
-
-[Desktop Action ExtractTo]
-Name=COSMIC Files Extract To...
-Exec=cosmic-files --extract-to %U
-NoDisplay=true
-
--- a/src/app.rs
+++ b/src/app.rs
@@ -791,9 +791,8 @@ impl App {
     fn open_file(&mut self, paths: &[impl AsRef<Path>]) -> Task<Message> {
         let mut tasks = Vec::new();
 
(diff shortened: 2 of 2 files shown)


jacobgkau:
I just tried downloading a ZIP file, and am seeing the same behavior with that (File Roller opening even though the default application is set to COSMIC Files in COSMIC Settings).

jacobgkau (commented):
Had some time to look at this again. Here's where we stand with this PR:

- The .desktop file lint is failing because the [spec](https://specifications.freedesktop.org/desktop-entry/latest/extra-actions.html) doesn't allow `NoDisplay` under individual actions. We can't simply ignore the lint because other distros have their own lints in packaging that would break if we did that. Moreover, the spec states "applications must only include actions that make sense as general launchers," so adding actions that would need to be hidden from the Applications menu wouldn't be the right approach, anyway.
    - Possible solutions I can think of would be using actions but implementing some kind of file-open GUI for them when launching without a filename (so they'd function if invoked from the Applications menu), or ditching the actions and instead special-casing the COSMIC Files options into COSMIC Settings.
- I'm still only getting a single `COSMIC Files` entry as a default application in Settings. So if we do stick with actions, Settings will need changing to support actions. (This could get messy with third-party apps' actions, so maybe it's not the best approach.)

All of that said, switching between COSMIC Files and other apps as the default does work as long as I restart COSMIC Files after changing the setting, as you pointed out.

Seeing as the current default behavior when double-clicking archives in COSMIC Files is the `Extract to...` dialog, and there's already no way to configure that, I think we could decouple the addition of allowing the direct `Extract` option as a default, and focus only on making the app itself configurable for now. So get rid of the actions from the .desktop file, and just make sure COSMIC Files is the default app by default, and that we show the `Extract to...` dialog when double-clicking archives (as we already do now) when that's the case.

PR #1564, opened 2026-01-25
Adds a way to copy the file path of a file (by adding it to the right click context menu)

orta:
There isn't really a way to get the full filepath for a file from the Files app. 

As an example alternative, on macOS, Apple make it so that the right click has a 'copy file name' which you can hold 'alt' to toggle into the full filepath. I'm not sure if that's too hidden, or something you want to copy - but I'm pretty sure this existed on the old desktop manager pre-COSMIC

Looks like:

image

( I built this with Claude Code, I'm not an expert Rust person, but I stand behind the code, it feels right judging by how the other context menu items work )

### DIFF
--- a/i18n/en/cosmic_files.ftl
+++ b/i18n/en/cosmic_files.ftl
@@ -337,6 +337,7 @@ quit = Quit
 edit = Edit
 cut = Cut
 copy = Copy
+copy-path = Copy file path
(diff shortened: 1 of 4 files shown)


orta:
I've happy to have the feature! No complaints from me, thanks for the note

Issue #1422, opened 2025-12-08
Support File Transfer portal for drag-and-drop and copy-paste to sandboxed Flatpaks


ids1024:
Drag-and-drop by path doesn't work if a sandboxed app doesn't have access to that part of the file system. For that purpose, apparently there's a "File Transfer" portal: https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.FileTransfer.html

We should be able to advertise the relevant mime time, then when it's used have some calls to [the `ashpd` bindings for File Transfer](https://docs.rs/ashpd/0.12.0/ashpd/documents/struct.FileTransfer.html) to transfer the file contents.

This will not, however, help with Chrome or Electron apps, since Chromium still doesn't support this portal: https://issues.chromium.org/issues/40898398

p-sims:
Will this allow you to drag a file to flapak app in dock and that will trigger the file to be opened with that application?

KodeBarista:
There's a issue for it here https://github.com/pop-os/cosmic-applets/issues/581

Issue #477, opened 2024-09-19
Drag-Drop Not working


DarkXero-dev:
Another one for you guys.

Dragging n Dropping is currently broken, like dragging a file from files to any app does nothing sometimes bugs out. I do apologize if this is not related to files maybe shell I dunno. Move this to where it belongs if in wrong place.

- Version/Distro
I am on **Arch** git version commit [**8668c18**](https://github.com/pop-os/cosmic-files/commit/8668c186072e508bb90286f3f02d70b0ed96e0dc)

Thanks

Elias-Graf:
I think it does work, it's just really, really, really (and I mean really) slow:

iamkartiknayak:
Drag & drop tested on latest version of 
- Chrome: works for both system & flatpak 
- Firefox & Zen: works on system not on flatpak

PR #1586, opened 2026-01-30
WIP File transfer portal

ids1024:
https://github.com/pop-os/cosmic-files/issues/1422

May as well put the work I did on this earlier in a draft PR. This should support copying using the portal, though that needs testing. We'll also want to support pasting using the portal.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1391,6 +1391,7 @@ name = "cosmic-files"
 version = "1.7.0"
 dependencies = [
  "anyhow",
+ "ashpd 0.12.3",
  "atomic_float",
  "bstr",
  "bzip2",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -397,6 +397,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "ashpd"
+version = "0.13.13"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fb8421aaa9644a5faf26735f258b669b15f063313ef8f8e2bdb28912a1a6f111"
+dependencies = [
+ "enumflags2",
+ "futures-util",
+ "getrandom 0.4.3",
+ "serde",
+ "serde_repr",
+ "tokio",
+ "zbus 5.18.0",
+]
+
+[[package]]
 name = "async-broadcast"
 version = "0.7.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1391,7 +1406,7 @@ name = "cosmic-files"
 version = "1.7.0"
 dependencies = [
  "anyhow",
- "ashpd 0.12.3",
+ "ashpd 0.13.13",
  "atomic_float",
  "bstr",
  "bzip2",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -67,7 +67,7 @@ thiserror = "2.0.18"
 atomic_float = "1.1.0"
 num_enum = "0.7.6"
 bstr = "1.12.1"
-ashpd = "0.12.0"
+ashpd = { version = "0.13.0", features = ["documents"] }
 
 # Completion-based IO runtime to enable io_uring / IOCP file IO support.
 [dependencies.compio]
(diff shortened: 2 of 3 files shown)


ItzExotical:
I gave this branch the testing it was waiting on. It works, including into a Flatpak browser, which is the case people keep reporting in #477.

### The key from this branch, retrieved inside a sandbox

Built `61e5e1e` and called `as_bytes("application/vnd.portal.filetransfer")` on a `ClipboardCopy`, then retrieved that key from inside Zen Browser's Flatpak sandbox with `gdbus`:

```
key from this branch: D_8_weTmaFpbSyGj_TyEVQ

$ flatpak run --command=sh app.zen_browser.zen -c 'gdbus call --session \
    --dest org.freedesktop.portal.Documents \
    --object-path /org/freedesktop/portal/documents \
    --method org.freedesktop.portal.FileTransfer.RetrieveFiles "$KEY" "{}"'
(['/run/user/1000/doc/h7ZUjRvzTkiehCpTyXAx9Q/dragged.txt'],)

zen reads: bytes the receiver should get
```

For contrast, the same browser being handed the plain path that `text/uri-list` carries today:

```
$ flatpak run --command=sh app.zen_browser.zen -c "head -c 60 /home/…/a-project/package.json"
head: cannot open '/home/…/a-project/package.json' for reading: No such file or directory
```

That is the whole bug in two lines. The browser holds `filesystems=xdg-download` and nothing else, so it can open a path under `~/Downloads` and no other. It is also why the reports read as intermittent: dragging from `~/Downloads` works and dragging the same file from anywhere else does not, which looks like a file manager quirk and is really the receiver's sandbox.

### On the `// XXX args`

`start_transfer(false, true)` is right for drag and drop. ashpd's signature is `start_transfer(writeable, auto_stop)`, so that is a read-only transfer stopped after the first `retrieve_files`, which is what GTK does for a drag and what you want here: the receiver is being handed a copy to read, and the transfer should not outlive the drop.

### On advertising both types

Both are correct. `strings` on GTK 4 and on Firefox's `libxul.so` shows each of them looking for exactly this pair:

```
application/vnd.portal.files
application/vnd.portal.filetransfer
```

Firefox also carries `FileTransferPortal::RetrieveFilesSync()` and the `org.freedesktop.portal.Documents` name, so the receiving half is already in place there. Nothing more is needed on the browser side.

### One review point

The two portal types are in the `LazyLock` list unconditionally, so they are offered even where no portal answers. A receiver that picks one then gets `None` from `as_bytes`, having passed over the `text/uri-list` it could have used. Advertising them only when the portal is reachable would keep the old behaviour intact on a machine without one. Not something I hit here, since the portal is present, but it is the one way this could make a working drag worse.

### Other notes

- Directories are fine in practice. ashpd's docs say only regular files may be added, but `AddFiles` accepted a directory fd and `RetrieveFiles` returned it, so a folder drag does not poison the transfer.
- The branch is 14 commits behind master and needs a rebase.
- I can push the round-trip test I used as a follow-up if that is welcome. It registers a file, retrieves it over a second connection the way a receiver would, and compares the bytes, and it skips itself where no portal is present so it will not fail CI.

### Environment

| | |
| --- | --- |
| Distribution | CachyOS (Arch-based) |
| Kernel | 7.2.6-1-cachyos, x86_64 |
| Desktop | COSMIC 1.8.0, Wayland session |
| cosmic-comp / cosmic-session | 1.8.0 |
| cosmic-files | this branch at `61e5e1e`; 1.8.0 installed from the distribution |
| Portals | xdg-desktop-portal 1.22.1, xdg-desktop-portal-cosmic 1.8.0, xdg-desktop-portal-gtk 1.15.3 |
| Receiver under test | Zen Browser 1.22.2b from Flathub, runtime org.freedesktop.Platform 25.08, `filesystems=xdg-download` only |
| CPU / GPU / RAM | Ryzen 9 7950X3D, Radeon RX 6750 XT (Navi 22), 64 GB |
| Toolchain | rustc 1.98.1 |

AI tools were used while investigating and writing this up. Everything above was run on the machine described: the build, the sandbox behaviour before and after, and the portal round trip using this branch's own code path.

PR #1594, opened 2026-02-05
feat: add setting to enable/disable the "recents" feature

### LINKED ISSUE #1591: [Feature request] Feature flag for recents
It would be nice if the user could disable the "recents" feature from the settings, this setting should also be saved to and loaded from the configuration.
This is how the file explorer works in other desktop environments and I always keep it disabled, I find it annoying when applications remember my activity, search history and similar.

I'm willing to open a PR for this feature when I have time, if this feature is wanted.

Krahos:
In this PR I added a toggle in the settings that allows the user to disable the "recents" feature.
This configuration is saved to and loaded from `~/.config/cosmic/com.system76.CosmicFiles/v1/` where it creates a new file, not sure if this is wanted/expected, but I saw also other settings are handled this way. It doesn't conflict with version 1.0.0 of cosmic files that I have installed in my system.

Closes #1591

### DIFF
--- a/i18n/en/cosmic_files.ftl
+++ b/i18n/en/cosmic_files.ftl
@@ -296,7 +296,7 @@ calculating = Calculating...
 ## Settings
 settings = Settings
 single-click = Single click to open
-show-recents = Show recents
+show-recents = Recents folder in the sidebar
 
 ### Appearance
 appearance = Appearance


nathansgithub:
This is a good feature to add. But does it make sense to make it a global setting to hide recent files in COSMIC apps/applets? Maybe there could be a way to hide "Recents" here separately, but people would probably mostly want that setting either on or off globally.

jacobgkau:
Maria is on the UX team, to be clear.

I don't think a less tech-savvy user right-clicking Recents and removing it accidentally is any more likely than doing the same with any of the other folders pinned by default. @maria-komarova Can you weigh in on if a right-click option should also be present, or if it should only be the switch in Settings?

maria-komarova:
> Isn't that too long? Especially for certain languages. But I'm neutral on the matter, I'll add a commit for this.

I don't think this is a concern for Settings since we have even longer labels in places. 

I also don't think we should add the option to the context menu (on right click). It might be difficult to know how to get this folder back if you remove it from the context menu since it is not going to show in the main view like other folders. So just the toggle in Settings.

jacobgkau (approved):
The label is now as Maria requested. The new setting's still working as expected. It's in sync between multiple windows if I open multiple before toggling the setting.

<details>

## Checklist

### Basic navigation

    - Not a regression.

### File operations

- [X] Files can be renamed.

- [X] Sorting options work.

### External filesystems

### Integrations

</details>

PR #1599, opened 2026-02-07
fix: disable dialog save button when filename is empty

### LINKED ISSUE #1576: [Enhancement] Disable Save button when filename is empty
**Issue/Description:**
Currently in the save dialog, Save button remains enabled even when no filename is provided. This allows users to click on the Save button that does nothing but animates a click, which is confusing and gives the impression that the action failed silently.

Greying out (disabling) the Save button until a filename is entered would improve UX and make the behavior consistent with the “Create New Folder” dialog, where the action button is disabled until a valid name is provided.

**Version:**
```
cosmic-files:
  Installed: 1.0.4~1769642128~24.04~e729ebe
```

**Steps to Reproduce:**
- Open any application that triggers the Save file dialog
- Leave the filename field empty
- Observe the state of the Save button

**Current Behavior:**

Image

**Expected Behavior:**
The Save button should be disabled when the filename field is empty and become enabled only after a valid filename is entered.

Image

**Other Notes:**
- Aligns behavior with the existing “Create New Folder” dialog

jasonrhansen:
Fixes #1576


jackpot51 (approved):

jackpot51:
@leviport in order to test this, it would have to have the cosmic-files crate updated in the cosmic-edit, cosmic-term, or xdg-desktop-portal Cargo.lock

PR #1646, opened 2026-02-20
feat: add up-directory button to navigation bar

### LINKED ISSUE #198: feature request: up arrow, to move up a directory
this is something that always annoyed my about most other file browsers, the lack of a up arrow, to move up to the parent directory.

the directory bar is a nice feature, and moving back and forth according to the path traversed is also handy, but sometimes it gets in the way, and you just want to move up the directory chain.

BlueKittyMeow:

Adds a navigate-to-parent button next to back/forward in the location bar. Reuses existing LocationUp message. Disabled at root. Closes #198.

### DIFF
--- a/src/tab.rs
+++ b/src/tab.rs
@@ -4926,7 +4926,7 @@ impl Tab {
 
         let size = self.size_opt.get().unwrap_or(Size::new(0.0, 0.0));
 
-        let mut row = widget::row::with_capacity(5)
+        let mut row = widget::row::with_capacity(6)
             .align_y(Alignment::Center)
             .padding([space_xxxs, 0]);
         let mut w = 0.0;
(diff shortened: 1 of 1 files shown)


jackpot51 (changes requested):
The PR template is mandatory, please fill it in.

jacobgkau:
This is what the button looks like in the navigation bar:

Screenshot_2026-03-11_10-57-02

@pop-os/ux Is this a feature we'd entertain, and does it need to be optional or require any other design work? The functionality itself is somewhat redundant with the breadcrumbs, but it is sort of a classic file browser feature.

jacobgkau:
As the UX team seemed not to be on-board with adding this button, and we didn't get a response from the PR author about how the current breadcrumbs don't address the same need, I'm going to close this PR for now.

We had some brief internal discussion about how we could make breadcrumbs more useful (e.g. make sure at least one directory up is always shown, so e.g. when shortening for the Home directory, don't shorten until the user navigates into a subfolder). We're still open to that sort of improvement going forward.

PR #1664, opened 2026-03-02
fix: navbar fills the available height of the window

hojjatabdollahi:
@wash2 the comment is not true. Extending the navbar all the way to the bottom works just fine.

____

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -2400,8 +2400,7 @@ impl Application for App {
         }
 
         Some(Element::from(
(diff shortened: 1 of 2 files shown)


garrettjwilke:
this builds and runs correctly...i am unsure how to test the differences of this fix vs what is already set

i have one system with this fix installed, and another system with the base `cosmic-files`...and functionally and visually they both appear the same...resizing them also yields the same results on both.

is there a way to explicitly show what this fixes?

thanks! :)

PR #1695, opened 2026-03-17
feat: repeat a key to cycle through items in type-to-select mode

sandr01d:
As discussed in #659, this PR extends the type-to-select mode introduced in #1471 with the ability to cycle through items starting with the same letter by repeatedly pressing a key. To make this work, type-to-select has been changed to start the search from the currently focused item (f467b9e).

- [X] I have disclosed use of any AI generated code in my commit messages.
- [X] I understand these changes in full and will be able to respond to review comments.
- [X] My change is accurately described in the commit message.
- [X] My contribution is tested and working as described.
- [X] I have read the [Developer Certificate of Origin](https://developercertificate.org/) and certify my contribution under its conditions.

### DIFF
--- a/src/tab.rs
+++ b/src/tab.rs
@@ -3029,29 +3029,43 @@ impl Tab {
             // considered first, otherwise last. Consider the focused item last when only a single character has been
             // typed, so we eagerly switch focus on the first character and stay on the same item as long as the prefix
             // matches.
-            let start = if prefix_lower.chars().count() == 1 {
+            let single_char = prefix_lower.chars().count() == 1;
+            let start = if single_char {
                 Self::index_after_focus(focus, self.sort_direction)
             } else {
                 Self::index_before_focus(focus, self.sort_direction)
             };
-            let Some((until, after)) = items.split_at_mut_checked(start) else {
-                log::error!(
-                    "invalid select focus index {start} for items of length {}",
-                    items.len()
-                );
-                return false;
-            };
-            let search_items = after
-                .into_iter()
-                .enumerate()
-                .map(|(i, item)| (i + start, item))
-                .chain(until.into_iter().enumerate());
+            self.select_focus = Self::select_first_prefix_from_index(
+                &prefix_lower,
+                items,
+                start,
+                self.sort_direction,
+            );
 
-            self.select_focus = if self.sort_direction {
-                Self::select_first_prefix_match(&prefix_lower, search_items)
-            } else {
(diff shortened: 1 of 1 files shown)


jacobgkau (changes requested):
This is a generally expected feature of type-to-select. It looks like starting at the currently selected item (if something's selected) is also how Dolphin handles it.

I am slightly concerned about the behavior when typing out the name of a file. Take the following test case:

- Create a folder
- In the folder, create these files:
    - 2025-08.txt
    - 2025-09.txt
    - 2025-10.txt
    - 2025-11.txt
    - 2025-12.txt
    - 2026-01.txt
    - 2026-02.txt

Say I want to select `2025-10.txt`. In Dolphin, when I type `2`, `2025-08.txt` is selected; as I continue typing, it remains selected until I get to the `-1`, at which point `2025-10.txt` is selected. This is also how COSMIC Files worked before this change.

On ad18050, when I type `2`, `2025-08.txt` is selected; when I type the `0`, `2025-09.txt` is selected; when I type `2`, `2025-10.txt` is selected; when I type `5`, `2025-11.txt` is selected. While this technically got me to what I wanted to select in less keystrokes, it easily leads to overshooting the desired file even though I'm still typing out its exact name. (And the feature generally encourages typing quickly, since there's a time delay on the typing.)

It would probably be better to retain the current behavior in general, and only cycle if the very first key is repeated. Dolphin does this, with the tradeoff that e.g. a `2200` file (sorted at the end in this example) would not get auto-selected since typing `22` would go to the first and then second file starting with `2`. It seems like that could be guarded against by only cycling on repeat if it otherwise would've led to no match. (If we do that, then maybe repeat cycling could still happen for all characters instead of just the first one.)

jacobgkau (approved):
This improvement's working as expected now.

With the same test folder, if nothing's selected, pressing `2` selects the first file; pressing `2` again cycles through all of the files starting with `2`. If I instead start typing a file name, such as `2026`, the selection jumps to that point in the folder. Moreover, if I add a file at the end called `2200-01.txt`, typing `2` twice selects that one first, then continues cycling on additional presses.

While further improvements might be possible in the future, this brings our cycling up to and beyond other implementations.

<details>

## Checklist

### Basic navigation

### File operations

- [X] Files can be renamed.

- [X] Sorting options work.

### External filesystems

### Integrations

</details>

wash2 (approved):

PR #1741, opened 2026-04-09
perf: fix slow startup by moving blocking trash checks to async background tasks

### LINKED ISSUE #1314: File Manager is slow to load when cifs mounted device is turned off
Related to https://github.com/pop-os/cosmic-files/issues/905.

If I have a cifs network mount and the remote device is turned off, Cosmic File Manager takes a long time to load.

trougnouf:
### Description
Fixes an issue where `cosmic-files` would take several minutes to start up or completely freeze<del>, spamming the console with `TrySendError { kind: Full }` warnings from `iced_futures::subscription::tracker` </del>.

### Cause
The `trash` crate's `is_empty()` and `trash_folders()` methods query all mounted filesystems. If the system has an unresponsive or slow network drive (NFS, SMB, or GVFS mounts), these synchronous calls block the thread they are running on. Because these methods were being called on the main UI thread (during menu/icon generation) and inside the subscription tracker, they prevented the application from rendering and blocked `iced` window events from being dispatched.

### Solution
* Created an `AtomicBool` (`TRASH_IS_EMPTY`) to act as an instantaneous cache for the UI to read from during synchronous menu and icon rendering.
* Shifted the `is_empty()` checks to `tokio::task::spawn_blocking` via a new `Message::UpdateTrashState`, triggered on app initialization and whenever `Message::RescanTrash` is dispatched. 
* Wrapped the `trash::os_limited::trash_folders()` call inside the `TrashWatcherSubscription` stream with a `tokio::task::spawn_blocking` block so it no longer locks up the `iced` event loop.

### Testing
- [x] Application starts instantly, even with unreachable network drives mounted.

Disclaimer:
* I used Gemini 3.1 Pro preview and Mistral devstral-2.
* There is a little bit of harmless unrelated stuff because I ran cargo clippy --fix.

I tested the changes; `cargo run --release` on https://github.com/pop-os/cosmic-files/commit/b17f8889a8a350a9a07ab3e00155d9b35f136152 results in a 2 minutes and 35 seconds start time (not counting build time). After these changes, `cosmic-files` start in < 1 second.

I've had this issue for months.
<del>This may be related to https://github.com/pop-os/libcosmic/issues/671</del>
It most likely  fixes #1314 

### DIFF
--- a/src/tab.rs
+++ b/src/tab.rs
@@ -39,7 +39,6 @@ use icu::{
 use image::{DynamicImage, ImageReader};
 use jiff_icu::ConvertFrom;
 use mime_guess::{Mime, mime};
-use regex::Regex;
 use rustc_hash::FxHashMap;
 use serde::{Deserialize, Serialize};
 use std::{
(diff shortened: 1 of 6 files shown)


jackpot51 (changes requested):
Please provide the clippy changes separately. The trash::os_limited functions and all the cfg surrounding them could be moved to another module. I'd vastly prefer this change to be done without an LLM

trougnouf:
> Please provide the clippy changes separately. The trash::os_limited functions and all the cfg surrounding them could be moved to another module. I'd vastly prefer this change to be done without an LLM

Done :) I `push -f` over the last commit s.t. I could get a clean diff without the clippy stuff, I hope that's ok. I don't plan on making another PR with the lint as anyone with write access can run `cargo clippy --all-features --all-targets --fix`, I just made sure none of the warnings were related to my changes.

mmstick:
@trougnouf The policy requires that you disclose which lines of code and commits contain code generated by AI in your commit messages. Copyright law also requires this because any code generated by AI is public domain and cannot be copyrighted.

trougnouf:
> @trougnouf The policy requires that you disclose which lines of code and commits contain code generated by AI in your commit messages. Copyright law also requires this because any code generated by AI is public domain and cannot be copyrighted.

Sorry about that. (I had written the original commit message before seeing the PR template and I thought it would be ok to disclose AI usage in the PR itself.) I've fixed the commit message now.

src/tab.rs
```diff
@@ -1161,133 +1159,7 @@ pub fn scan_search<F: Fn(SearchItem) -> bool + Sync>(
...
-                return;
-            }
-        };
-
-        for entry in entries {
-            if let Ok(metadata) = trash::os_limited::metadata(&entry).inspect_err(|err| {
-                log::warn!("failed to get metadata for trash item {entry:?}: {err}")
-            }) {
-                let name = entry.name.to_string_lossy();
-                if regex.is_match(&name) && !callback(SearchItem::Trash(entry, metadata)) {
```
mmstick, line 1287: Refactors would be better in their own separate commit.

src/trash_helpers.rs
```diff
@@ -0,0 +1,161 @@
...
+
+pub static TRASH_IS_EMPTY: AtomicBool = AtomicBool::new(true);
+
+pub fn is_empty() -> bool {
+    TRASH_IS_EMPTY.load(Ordering::Relaxed)
+}
+
+pub fn set_empty(is_empty: bool) {
+    TRASH_IS_EMPTY.store(is_empty, Ordering::Relaxed);
+}
```
mmstick: As far as I can tell, these are redundant. The value can be stored in the app struct.

trougnouf:
Done, thank you for reviewing.

When testing the trash's behavior it's far from perfect; opening it yields an empty directory until it is populated which can take minutes with no feedback. This is the same behavior that previously occurred in b17f8889a8a350a9a07ab3e00155d9b35f136152 , the only difference I noticed after e9ad9ac81a35827383e317b9b301aff07a6b2e2c (other than the instantaneous startup time) is that deleting something would previously freeze up the browser and it's now instantaneous.

mmstick:
There are git merge conflicts with this PR.

trougnouf:
> There are git merge conflicts with this PR.

@mmstick Done, please review.

There seem to be some issues with the `trash-rs` crate. I got the following messages:

```
thread 'tokio-rt-worker' (3430055) panicked at /home/trougnouf/.cargo/git/checkouts/trash-rs-9826275beb9646ce/a225f75/src/freedesktop.rs:292:5:
assertion failed: virtually_exists(&file).map_err(|e| fs_error(&file, e))?
[2026-04-27T19:52:26Z WARN  cosmic_files::app] failed to rescan: task 314 panicked with message "assertion failed: virtually_exists(&file).map_err(|e| fs_error(&file, e))?"

thread 'tokio-rt-worker' (3429656) panicked at /home/trougnouf/.cargo/git/checkouts/trash-rs-9826275beb9646ce/a225f75/src/freedesktop.rs:292:5:
assertion failed: virtually_exists(&file).map_err(|e| fs_error(&file, e))?
```

and the trash directory fails to populate. At least there is no crash/freeze in the app because it's on a separate async thread. Those errors go away and the trash behaves correctly if I delete the offending files.

ArturRuta:
# The issue comes from both using the main thread and repeating the searches for .trash continuously

I have a similar issue open as #1986 that i believe is worth reviewing together with this one since
- I've collected relevant trace info that pintpoints the behavior
- the underlying problem is **not only** that the search for trash folders is done in the main thread for remote filesystems and tehefore generates freezes of the UI. The probles comes also from the fact that this search **is repeated time after time**, more than once per second in fact. 

My concern is that if we just move the search for trash files to a different thread we will restre the gui responsiveness but still cosmic-files will be
- Checking **remote** filesystems for trash-folders (which I'm not sure is apropiate) and
- Doing it **repeatedly no matter the result of the search just done** (which i definitely think is not a proper behavior)  

Sorry if I'm mislead, i just ran into the problem amb was unsure that this issue was going to cover the whole scope of the problem.
I've open my findings on a separate issue so that you can decide whether they are relevant or not without "trashing" this issue.
On the other side since i can reproduce at will I would be glad to collect any trace or additional information you may need i that helps.
Best regards

PR #1747, opened 2026-04-12
fix: progress bar caps at 50% for cross-device moves (MTP/external drives)

### LINKED ISSUE #1457: [Bug] File move progress completes at 50%
**Issue/Description:**
Currently, when moving files to a different drive, the progress indicator often finishes at `50%` instead of reaching `100%`, even though the operation completes successfully. This suggests an issue with how progress percentage is calculated or reported. The UI gives the impression that the operation finished halfway, which is misleading and confusing.

**Version:**
```
cosmic-files:
  Installed: 0.1.0~1766186873~24.04~84d3659
```

**Steps to Reproduce:**
- Open COSMIC Files
- Mount an external or secondary drive
- Move a file/folder to a different drive
- Observe the progress indicator
- Notice the operation completes when progress reaches ~`50%`

**Expected Behavior:**
File move progress should accurately reflect completion and reach `100%` when the operation finishes.

Cj1m:
Closes #1457

## Summary

When moving files to an external device (MTP, cross-device), the progress bar stalls at 50% and then jumps straight to 100% when the transfer completes.

## Root cause

\`recursive_copy_or_move\` builds an \`ops\` list containing both the transfer op and a cleanup \`Remove\`/\`Rmdir\` op (to delete the source after copying). \`total_ops\` was set to \`ops.len()\`, so the progress formula:

\`\`\`
(item_progress + current_ops) / total_ops
\`\`\`

could only reach \`(1.0 + 0) / 2 = 0.5\` for a single-file move — cleanup ops inflate the denominator but carry no byte-level progress themselves.

## Fix

- Exclude cleanup ops from \`total_ops\` — they are fast file deletions, not data transfers
- Use a manual \`current_ops\` counter that only increments for non-cleanup ops

## Test

Added \`operation::tests::move_progress_does_not_cap_at_50_percent\` which calls \`Context::recursive_copy_or_move\` directly with \`Method::Move\`, captures all progress values from the \`on_progress\` callback, and asserts the maximum reaches \`1.0\`.

---


jackpot51 (changes requested):
The PR template is mandatory

Cj1m:
@jackpot51 Apologies, missed this. I have updated accordingly.

Issue #1703, opened 2026-03-22
[Feature Request] Allow Customization of Context Menu Entries


KodeBarista:
**Issue/Description:**
COSMIC Files currently does not provide an option to customize or configure context menu entries. For my use case actions such as `Copy to` and `Move to` are always visible, even if they are not used frequently. This results in unnecessary clutter in the context menu and reduces efficiency for users who prefer a more minimal or tailored set of actions.

**Version:**
`cosmic-files-1.0.8-1.fc43.src.rpm`

**Expected Behavior:**
Users should be able to configure which context menu entries are visible, allowing them to hide unused actions, in this case entries such as `Copy to` and `Move to`.

darkfated:
You can add your own options #1728. Regarding hiding unnecessary things, I don't know if this is a good idea

neogeographica:
I've tried adding options as per #1728 and I haven't yet got it to work.

For example I currently have this file at ~/.config/cosmic/com.system76.CosmicFiles/v1/context_actions :
```
[
    (
        name: "Open in ghostty",
        confirm: false,
        selection: Folders,
        steps: [
            "ghostty +new-window --working-directory=%F",
        ],
    ),
    (
        name: "Debug",
        confirm: false,
        selection: Any,
        steps: [
            "echo hi >> /tmp/debug.log",
        ],
    ),
]
```

The appropriate new context menu options do appear when I right click a file or folder selection, but selecting the option has no apparent result -- doesn't launch ghostty or write to /tmp/debug.log .

If I set confirm to true, I do get the confirmation dialog, but again no result.

Is there a place I can look for any error messages or other output that might be getting generated here?

edit: Possibly interesting: if I change the "Debug" command to this: `sh -c 'echo hi >> /tmp/debug.log'` then it does work. Will experiment a bit more there.

neogeographica:
OK well, hmm. The reason the ghostty command is not working is that the "steps" parsing uses shlex.split and then looks for individual '%F' strings in the resulting list. However in this case the list element was the whole chunk "--working-directory=%F" so it didn't do any substitution for the %F. This caused the ghostty launch to fail.

Unfortunately that's the syntax that ghostty requires to do this, as far as I can tell. If the parsing is going to be that rigid I guess I can get around it with a little wrapper script that constructs the correct command line for ghostty.

Edit: yep.

Script "gdir.sh" in my PATH:
```sh
#!/usr/bin/env sh
ghostty +new-window --working-directory=$1
```

[...]

PR #1775, opened 2026-04-24
fix: improve desktop entry field code handling

mmstick:
- The %f and %u field codes may now be expanded within a word
- Handle field code escapes (%%)
- Support the %c and %k field codes

This will notably fix desktop entries and context menu actions that pass files as a long argument, such as `--option=%f`.

I will be moving this into libcosmic soon as well.

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1375,6 +1375,7 @@ version = "1.0.11"
 dependencies = [
  "anyhow",
  "atomic_float",
+ "bstr",
  "bzip2",
  "compio",
  "cosmic-client-toolkit",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,6 +65,7 @@ tracing-subscriber = { version = "0.3.23", features = ["env-filter"] }
 thiserror = "2.0.18"
 atomic_float = "1.1.0"
 num_enum = "0.7.6"
+bstr = "1.12.1"
 
 # Completion-based IO runtime to enable io_uring / IOCP file IO support.
 [dependencies.compio]
--- a/src/app.rs
+++ b/src/app.rs
@@ -884,31 +884,39 @@ impl App {
     #[cfg(feature = "desktop")]
     fn launch_desktop_entries(paths: &[impl AsRef<Path>]) {
         use cosmic::desktop::fde::DesktopEntry;
+        let locales = cosmic::desktop::fde::get_languages_from_env();
 
         for path in paths.iter().map(AsRef::as_ref) {
             match DesktopEntry::from_path::<&str>(path, None) {
                 Ok(entry) => match entry.exec() {
-                    Some(exec) => match mime_app::exec_to_command(exec, &[] as &[&str; 0]) {
-                        Some(commands) => {
-                            let cwd_opt = entry.desktop_entry("Path");
+                    Some(exec) => {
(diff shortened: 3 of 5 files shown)


jacobgkau (commented):
This doesn't make the ghostty example given in https://github.com/pop-os/cosmic-files/issues/1703#issuecomment-4315131721 work. It still fails to launch with `--working-directory=%F` in the context action. Is this supposed to work?

jacobgkau:
Oh, it works with `--working-directory=%f` (lowercase F).

jacobgkau (changes requested):
An action with `cosmic-term -w %F` used to work. It now needs to be `cosmic-term -w %f`. That seems like a regression, unless it's an intentional breaking change?

mmstick:
%F isn't permitted to be used with adjacent characters touching it since it expands into multiple separate arguments. The first one needs to use %f since that's for substituting one path per command.

Using %f therefore makes multiple commands with one path per command while %F combines all paths into arguments for a single command that can accept multiple file arguments.

mmstick:
For reference https://specifications.freedesktop.org/desktop-entry/latest/exec-variables.html

darkfated:
Good pull request

| Placeholder | Description | Example |
|-------------|-------------|---------|
| `%%` | A literal percent sign | `%%` -> `%` |
| `%c` | Application name (entry name) | `%c` -> `"Files"` |
| `%k` | Path to the `.desktop` file | `%k` -> `"/usr/share/applications/org.pop_os.CosmicFiles.desktop"` |
| `%f` | A single selected file. For multiple files, the command runs separately for each | `myedit %f` -> `myedit file1.txt` (then `myedit file2.txt`) |
| `%F` | **All selected files** - each passed as a separate argument **in a single command invocation** | `code %F` -> `code file1.txt file2.txt folder/` (all in one window) |

jacobgkau (commented):
@mmstick Your explanation makes sense for why it won't work with `=` syntax, but I still don't understand why the cosmic-term example breaks. While cosmic-term doesn't handle multiple paths properly, if I right-click only a single directory, it would still be only one path for `%F` to expand out to, and that should still result in a valid cosmic-term command, right?

Also, aside from custom context menu actions, it looks like even the built-in `Open in terminal` option doesn't work on this branch (5e8cbb4 from before and d775f3e from today), which I hadn't noticed last week. It works on master (d5dbcc7).

mmstick:
Yes, if there's only one path it would create one command that behaves the same regardless of %f or %F.

jacobgkau (approved):
`cosmic-term -w %F` now works again. Ghostty with `--working-directory=%F` still doesn't, but `--working-directory=%f` still does and `--working-directory %F` does as well (with the former being more correct per your explanations). The built-in `Open in terminal` option is also fixed. `cosmic-edit %F` still works correctly to open multiple text files in separate tabs in the same instance of COSMIC Edit (while `cosmic-edit %f` opens them in separate instances/windows).

<details>

## Checklist

### Basic navigation

### File operations

- [X] Files can be renamed.

    - Some text appears too light (low contrast) in light theme, not a regression: https://github.com/pop-os/cosmic-files/issues/1786 
- [X] Sorting options work.

### External filesystems

### Integrations

</details>

jackpot51 (approved):

PR #1801, opened 2026-05-10
fix(i18n): remove ellipsis from rename action labels

### LINKED ISSUE #1762: Remove "..." from Rename button in the Rename dialog
We use "..." to indicate that interacting with an item will require more actions from the user. This is not the case with Rename dialog where pressing "Rename" will result in a single complete action.

KodeBarista:
- [X] I have disclosed use of any AI generated code in my commit messages.
- [X] I understand these changes in full and will be able to respond to review comments.
- [X] My change is accurately described in the commit message.
- [X] My contribution is tested and working as described.
- [X] I have read the [Developer Certificate of Origin](https://developercertificate.org/) and certify my contribution under its conditions.

closes https://github.com/pop-os/cosmic-files/issues/1762


KodeBarista:
I'm not sure whether this can be merged from here, since the project started using Weblate. Fixing this required updating the translations across all languages, which would otherwise need to be handled by the respective reviewers. If this causes issues with Weblate, please feel free to close the pull request.

jacobgkau:
I think it might need to be done here for the English base, but probably needs to be done on Weblate for the rest of the languages.

KodeBarista:
Then perhaps it might be better to do it in weblate for all languages? I can share the link of the issue in translation channel for bringing it to the reviewers attention.

mmstick:
It's best to only update the English keys. Changing other files will cause repository conflicts. Though do note that any key that is modified in the English file will cause all pending translations of that key to be automatically deleted from Weblate when it does a repository sync.

mmstick:
#1745 is a better fix since it adds a new key for the dialog instead of modifying an existing one.

lorduskordus:
I just wanted to say the 'rename' key is used in two different cases, the ellipsis is desired in one and not in the other, so this would break one of the cases, but you already noticed my PR, I explained it in more detail there

PR #1810, opened 2026-05-14
feat: show absolute path in details view

nz366:

### DIFF
--- a/src/tab.rs
+++ b/src/tab.rs
@@ -1756,6 +1756,8 @@ pub enum Message {
     HighlightActivate(usize),
     DirectorySize(PathBuf, DirSize),
     ImageDecoded(PathBuf, u32, u32, Vec<u8>, Option<(u32, u32)>, u64), // path, width, height, pixels, display_size, generation
+    CopyPath,
+    Ignore
 }
 
 #[derive(Copy, Clone, Debug, Eq, PartialEq)]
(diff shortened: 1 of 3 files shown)


src/tab.rs
```diff
@@ -2337,6 +2339,16 @@ impl Item {
 
         let mut details = widget::column::with_capacity(8).spacing(space_xxxs);
         details = details.push(widget::text::heading(self.name.clone()));
+        if let Some(path) = self.path_opt() {
+            details = details.push(
+                widget::row![
+                    widget::text_input("", path.to_string_lossy()).on_input(|_| Message::Ignore),
```
mmstick: Is there a reason to emit a message if the message is ignored?

nz366: on_input requires a Message to be returned. without this method iced shows a text field that can't be selected.

jklgrasso:
@pop-os/ux This change uses the 'text_input' widget. This causes a cursor to appear. Upon receiving input, text is not written, but the cursor moves.

Current:
image


Proposed:
image

nz366:
~~Will have to wait until proper widget arrives https://github.com/iced-rs/iced/pull/3315~~

jklgrasso:
That looks better to me. I'm not a huge fan of how the text box looks here, but that's up to UX.

The 'Cargo.toml' can't change like that. If a dependency needs to be modified for this change to work as you intend, a PR needs to be made/merged in that dependency's repo.

nz366:
> PR needs to be made/merged in that dependency's repo.

Yeah, I will start that now

maria-komarova:
Sorry for the delay. I had a design for the updated layout for File details in the works, wanted to make sure that one was ready to go because it has Copy path button. Please see the design here: #1848. There is more to the new design than just the button so not sure if the button should be added in this PR or as part of the larger adjustment.

nz366:
Should I change the scope of this pull request for this updated design?

maria-komarova:
If you would like to work on it that would be a good contribution.

nz366:
There's also translation templates that needs to be updated. on a second thought I should leave this to the maintainers.

PR #1836, opened 2026-06-06
Fix: Space Key Overrides Type to Search Setting

### LINKED ISSUE #1831: [Bug] Not possible to use space when typing to select file
**Description**

With "Type to search" set to "Selects the first matching file or folder", it's not possible to jump to a file with space in its filename, since space triggers the file manager's pop-up preview.

It would be nice if space was properly mapped to the typeahead input while in typeahead mode.

<details>
<summary>Video evidence</summary>

https://github.com/user-attachments/assets/6303cb92-bbcb-4887-adad-31688878b348

</details>

**Environment**

COSMIC FIles: `1.0.14~1779995384~24.04~bf01bb3`

LukasBauza:
Fixes: #1831

When "Type to search" is set to "Selects the first matching file or folder", it's not possible to type out the name of a file that includes a Space character within the filename, as it causes the Gallery to pop up.

Changed the shortcut for the Gallery to pop up from Space to Shift+Space.

AI Usage: Used Gemini to help me figure out how to navigate the code. No code was generated by AI for the commit.

### DIFF
--- a/src/key_bind.rs
+++ b/src/key_bind.rs
@@ -39,7 +39,7 @@ pub fn key_binds(mode: &tab::Mode) -> HashMap<KeyBind, Action> {
     bind!([Ctrl, Shift], Key::Character("n".into()), NewFolder);
     bind!([], Key::Named(Named::Enter), Open);
     bind!([Ctrl], Key::Character(" ".into()), Preview);
-    bind!([], Key::Character(" ".into()), Gallery);
+    bind!([Shift], Key::Character(" ".into()), Gallery);
 
     bind!([Ctrl], Key::Character("h".into()), ToggleShowHidden);
     bind!([Ctrl], Key::Character("a".into()), SelectAll);


wash2 (approved):

jklgrasso:
I can see this change causing some initial confusion for those who are used to `space` for previews.

Could it be better to have an option for choosing between `space` and `shift` + `space`? Or when using the "Selects the first matching file or folder" option, the keybind for file preview changes to `shift` + `space` while keeping the current preview keybind for other "Type to search" options?

If we are okay with this change as is, I will approve it. Just thought I'd ask beforehand.

LukasBauza:
> I can see this change causing some initial confusion for those who are used to `space` for previews.
> 
> Could it be better to have an option for choosing between `space` and `shift` + `space`? Or when using the "Selects the first matching file or folder" option, the keybind for file preview changes to `shift` + `space` while keeping the current preview keybind for other "Type to search" options?
> 
> If we are okay with this change as is, I will approve it. Just thought I'd ask beforehand.

Yeah, that is a fair thing to point out, as `space` is the default key binding.

The issue I see is if the shortcut for previews is changed based on what option is chosen within "Type to search", the user would have to know that the shortcut changes between `space` and `shift` + `space` based on what the option is, and would have to remember the change. If the user changed between the "Type to search" option frequently, this can feel confusing.

Having a choice between what key binding they can use feels more of a feature in allowing the users to change the key bindings for the application, as I think this can be applied to any key bind.

LukasBauza:
Still waiting for a review on this.

Issue #70, opened 2024-03-07
[Feature Request] Resizable left pane


gantryyork:
It would be nice is the left pane were resizable

lsqiu1981:
I've been looking forward to this feature for a while now.

PR #1872, opened 2026-06-20
feat: Resizable left pane

ZlordHUN:
Implements https://github.com/pop-os/cosmic-files/issues/70

- Saves resized values to config so the resizing is preserved. Double clicking the edge of the left panel restores the default size.

[Kooha-2026-06-20-14-55-45.webm](https://github.com/user-attachments/assets/ed7e34bd-068b-4f30-a047-7291a33ede70)


jacobgkau (changes requested):
In the below screenshot, my network mount's text is cut off mid-character:

Screenshot_2026-07-23_16-26-25

I think we'd want the text to be ellipsised instead of simply cut off.

jacobgkau:
This would need approval from @pop-os/ux. It also seems like something we'd want to apply to all apps with a similar navigation bar, rather than just COSMIC Files, if we wanted to support this (not sure if that would raise any additional engineering implications).

ZlordHUN:
@jacobgkau Done. While testing after the rebase, I found and fixed a bug where reversing the pointer during a resize could cause the inferred drag direction to flip between frames, preventing smooth resizing.

Issue #1766, opened 2026-04-17
Design for showing the count and size of the selected files


maria-komarova:
The issue has come up multiple times. The designs below demonstrate how we'd like to show the count and size of the selected items. It goes a little beyond that by including a few useful actions along with this information. There are two ways:
1. When File details drawer is open
2. When File details drawer is closed.
[Figma link.](https://www.figma.com/design/ZBg577prx1KF8BlZrBSgT0/Files-Handoff?node-id=5998-36755&m=dev)
<img width="4189" height="2482" alt="Image" src="https://github.com/user-attachments/assets/e1b0154a-a29b-46ee-afb3-299b7cd7f073" />

jacobgkau:
In discussion with Maria about the footer, it sounds like it should actually not be something that floats on top of content as depicted in the first mockup, but instead should be more of a panel that shrinks the scrollable view above it (to avoid a situation where you can't fully view the bottom row of a folder).

PR #1754, opened 2026-04-14
feat(files): show count and total size of selected files in bottom bar (issue #1683)

ChonkyMonkey:
Alright, this is an edit of the original message because the inital design got quite an overhaul.

This PR implements the file size + counter as described in issue #1683 and with additional UI/UX suggestions from #1766.

### Key Features
- when items are selected a floating footer appears, which shows file number and size as well as action buttons
-- the available actions are `Select all`, `Delete`, `Open`, `Duplicate` and `Move to...`
- when the file details tab is open, these elements appear in the right panel
- when inside the trash folder the options change to `Select all`, `Delete permanently`, `Restore from trash` and `Move to...`

### Behaviour Details
- when all items are selected the `Select all` button changes to a `Deselect all` button
- using the `Move to...` action from trash will restore the items in the specified directory
- selecting only a single file will display the old details tab; selecting 2 or more will display the new details tab

### Image gallary
Screenshot_2026-04-23_00-07-51
Screenshot_2026-04-23_00-08-09
Screenshot_2026-04-23_00-08-23
Screenshot_2026-04-23_00-11-17
Screenshot_2026-04-23_00-08-39
Screenshot_2026-04-23_00-09-03

A couple of notes:
- The floating footer is overlaid on top of the tab view, and the tab content gets a bottom inset when that footer is present. This was done to keep the bottom most content visible even if the footer is active.
- The destination dialog flow is shared between normal move actions and move-from-trash via `FileDialogContext`, rather than adding a separate dialog/result path for trashed items.

I'm not 100% happy with the current details page implementation. When only one item is selected I chose to show the old page to avoid dealing with coming up with a new UI. #1766 unfortunatly omitted the file details display (type, size, created, modified, accessed) in their design, so I couldn't take theirs 1 for 1. 
This causes a huge UI shift when selecting 1 item vs selecting multiple, which should be changed at a later stage I guess.

---

- [x] I have disclosed use of any AI generated code in my commit messages.
  - If you are using an LLM, and do not fully understand the changes it is making to the code base, do not create a PR.
  - In our experience, AI generated code often results in overly complex code that lacks enough context for a proper fix or feature inclusion. This results in considerably longer code reviews. Due to this, AI authored or partially authored PRs may be closed without comment.
- [x] I understand these changes in full and will be able to respond to review comments.
- [x] My change is accurately described in the commit message.
- [x] My contribution is tested and working as described.
[...]

PR #1874, opened 2026-06-20
improv: make type-to-select prefix editable and visible

lihe07:
A screenshot preview of the new prefix search chip (bottom right)

- Backspace now edits the active prefix and Escape clears it, so fixing a typo or starting a fresh search no longer requires waiting for the timeout.
- Show a bottom-right floating prefix indicator while filtering.
- Auto-clear the buffer (hiding the indicator) once it idle, via a subscription.

Applies to both the main window and the open/save dialog.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -2782,6 +2782,12 @@ impl Application for App {
                 return Task::none();
             }
 
+            if !self.type_select_prefix.is_empty() {
+                self.type_select_prefix.clear();
+                self.type_select_last_key = None;
+                return Task::none();
+            }
+
             let had_focused_button = tab.select_focus_id().is_some();
             if tab.select_none() {
                 if had_focused_button {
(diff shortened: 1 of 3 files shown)


jacobgkau (commented):
@pop-os/ux would need to approve this feature for it to be accepted, so pinging them. For reference, Caja and Nemo both have an overlay like this, but they don't support cycling through results with repetition of the first key like we currently do. (Dolphin supports repetition, but doesn't have an overlay.)

One thing I noticed quickly: When repeating the first-typed character to cycle through everything starting with that character, the letter shows up multiple times in the new overlay. Additionally, pressing Backspace continues cycling forward through items (which doesn't happen in the currently released version of COSMIC Files):

https://github.com/user-attachments/assets/d8a14fd4-b303-4445-9d75-43d605d64d32

Since we prioritize exact matches over repetition (see https://github.com/pop-os/cosmic-files/pull/1695 for details), it seems like showing the entire string that's entered does make sense. However, it seems like it'd be more useful for Backspace to cycle back up the list, instead of continuing down. Not sure how easy of a change that would be.

An alternative approach would be to only show a letter once in the pop-up if it's being repeated to cycle through items, but if we did that, then I'm not sure how we'd handle the case where e.g. 2 presses selects an exact match (which means the second letter should show up in the overlay) but then a third press goes back to cycling through.

lihe07:
I agree with the backspace tweak, and I added a previous selection stack so Backspace will pop it and restore the last selection.

https://github.com/user-attachments/assets/ede0d96d-54a0-47c0-b57e-44b8feb74bf6

jacobgkau:
After speaking with our UX team, it sounds like we do have room for this feature in COSMIC Files, but we'd like it to be integrated into the same footer bar that we're planning to use for showing selected file count and size: https://github.com/pop-os/cosmic-files/issues/1766 @maria-komarova will have updated designs that include the two features together soon.

We had https://github.com/pop-os/cosmic-files/pull/1754 open to implement the footer bar for selected file count and size, but our engineering team didn't like it due to it being written suboptimally by AI, so we're currently waiting for either someone else to redo that or for our engineering team to have time to do it themselves.

maria-komarova:
The designs are shown [here in Figma](https://www.figma.com/design/ZBg577prx1KF8BlZrBSgT0/Files-Handoff?node-id=6150-6603&m=dev). 
The idea is to use the same footer and Context Drawer we use for showing the information about the number of selected items to show the status of the typeahead. The video below shows me typing "e" in the prototype, followed by another "e", then pressing Backspace two times, then typing "e" again. After that I am selecting another item which switches the information back to the number of the selected items. The example of the Context Drawer with File details is shown in Figma. 
 
https://github.com/user-attachments/assets/574669eb-e0c6-498e-a95d-3fefa9f99c4d

lihe07:
Appreciated the inputs! I will propose an implementation of the footer.

KodeBarista:
Any updates?

PR #1905, opened 2026-07-03
Fix folder named `files` being shown as empty

### LINKED ISSUE #1903: Folder literally named `files` is shown as "Empty folder"
**Describe the bug**

Any directory whose name is exactly `files` (lowercase) is shown as **"Empty folder"** in COSMIC Files, even when it contains items. The parent view still shows the correct item count for that folder (e.g. "3 files"), so it only appears empty once you open it.

The cause is an over-broad trash check: `is_trash_path` (`src/trash.rs`) treats a path as being inside the Trash whenever **any** ancestor directory is named `files`, not just the real XDG Trash `files/` directory. COSMIC Files then tries to reconstruct each child as a trashed item, which fails for a normal folder, so every child is silently dropped.

```rust
// src/trash.rs
pub fn is_trash_path(path: &Path) -> bool {
    path.ancestors().any(|a| a.ends_with("files")) // matches ANY folder named "files"
}
```

In `scan_path()` (`src/tab.rs`), each entry of such a folder is routed through `item_from_trash_child` → `original_path_for_trash_child`, which reads a sibling `info/<name>.trashinfo` that does not exist, returns `None`, and the enclosing `filter_map` drops the entry — hence an empty listing.

The match is whole-component and case-sensitive, so it also affects e.g. `/var/www/html/files` and any descendant of a `files/` directory. `Files`, `FILES`, `files2` are unaffected.

**To Reproduce**
1. `mkdir -p ~/tmp/cosmo-test/files`
2. `touch ~/tmp/cosmo-test/files/{a.txt,b.csv,c.jpg}`
3. Open `~/tmp/cosmo-test` in COSMIC Files — the `files` folder correctly shows "3 files".
4. Double-click into the `files` folder.
5. It is shown as "Empty folder".

For contrast, sibling folders named `Files`, `FILES`, `files2`, or `notfiles` (each with 3 files) all display their contents correctly — only the exact lowercase `files` folder appears empty.

**Expected behavior**
A folder named `files` lists its contents like any other folder. Only the real XDG Trash `files/` directory should be treated as trash.

**Desktop (please complete the following information):**
 - OS: Pop!_OS 24.04 LTS (COSMIC, Wayland)
 - cosmic-files 1.2.0 (pkg `1.2.0~1782846793~24.04~3ef7913`, commit 3ef7913). Still present on `master` (`src/trash.rs`), introduced by 4910718 "feat: view folder's content in trash".

**Additional context**
A fix PR follows: restrict `is_trash_path` to a `files/` directory that also has the sibling `info/` directory mandated by the FreeDesktop trash spec, so ordinary folders named `files` are no longer mistaken for the Trash.

chris-010:
## Summary

Fixes #1903.

A folder literally named `files` was displayed as **"Empty folder"** in COSMIC Files, even when it had contents (the parent view still showed the correct item count). `is_trash_path` treated any path with an ancestor component named `files` as being inside the Trash, so every child of such a folder was routed through the trash-item reconstruction path (`item_from_trash_child` → `original_path_for_trash_child`), which returns `None` for entries that have no `.trashinfo` metadata — emptying the listing.

This affects any directory named exactly `files` and all of its descendants (e.g. `/var/www/html/files`), not just top-level folders. Introduced by 4910718 ("feat: view folder's content in trash").

The fix restricts `is_trash_path` to a `files/` directory that also has the sibling `info/` directory mandated by the [FreeDesktop trash specification](https://specifications.freedesktop.org/trash-spec/latest/), so an ordinary folder named `files` is no longer mistaken for the Trash while the real Trash (which always has both `files/` and `info/`) still is.

**Testing:** added unit tests in `src/trash.rs` covering the regression (a plain folder named `files` is not trash), a real trash layout (`files/` + `info/` is trash, including nested paths), and the guard where a sibling `info` is a regular file. `cargo test --lib trash::` passes.

---

### DIFF
--- a/src/trash.rs
+++ b/src/trash.rs
@@ -105,7 +105,15 @@ pub fn original_path_for_trash_child(p: &Path) -> Option<PathBuf> {
 
 /// Check whether a path is inside any trash `files/` directory.
 pub fn is_trash_path(path: &Path) -> bool {
-    path.ancestors().any(|a| a.ends_with("files"))
+    path.ancestors().any(is_trash_files_dir)
+}
+
+/// Whether `dir` is the `files/` directory of an XDG trash bin: it is named
+/// `files` and has the sibling `info/` directory that the FreeDesktop trash
+/// specification requires. This distinguishes a real trash bin from an ordinary
+/// user folder that merely happens to be named `files`.
+fn is_trash_files_dir(dir: &Path) -> bool {
+    dir.ends_with("files") && dir.parent().is_some_and(|bin| bin.join("info").is_dir())
 }
 
 pub struct Trash;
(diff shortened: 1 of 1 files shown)


jackpot51:
This is an improvement, but it should probably test if the higher directory is Trash or .Trash

KodeBarista:
Looks like this has been already addressed. I'd the this issue yesterday but after 1.4.0 [update](https://github.com/pop-os/cosmic-files/commit/9a21542825af92811b622a58ace5f25f962506d7), it went away.

image

chris-010:
This was fixed independently by #1933 (merged in 1.4.0), using essentially the same approach — a `files/` directory only counts as trash if its parent is an actual trash root. I've confirmed on my system after updating to `1.4.0~…~9a21542`: the affected folder shows its contents again and the trash view still works. Closing as superseded — thanks @KodeBarista for the confirmation.

PR #1910, opened 2026-07-08
Fix "open archive" button

krakotay:

#be67fd8e This old commit made 'Open' (archive) behave like 'Extract To', which is very strange, duplicates behavior, and is just confusing. Restored the expected behavior
image

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -791,23 +791,13 @@ impl App {
         // This allows handling paths as groups if possible, such as launching a single video
         // player that is passed every path.
         let mut groups: FxHashMap<Mime, Vec<PathBuf>> = FxHashMap::default();
-        let mut all_archives = true;
-        let supported_archive_types = crate::archive::SUPPORTED_ARCHIVE_TYPES;
         for (mime, path) in paths.iter().map(|path| {
             (
                 mime_icon::mime_for_path(path, None, false),
                 path.as_ref().to_owned(),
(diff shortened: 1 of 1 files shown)


mmstick (approved):

jacobgkau:
This would need UX approval, as I'm not sure it's the intended solution.

Right-clicking and clicking Open should generally do the same thing as double-clicking an item. It's currently intentional that double-clicking an item opens the `Extract to folder` dialog.

We would like this to be configurable so users can set a different default archive app if they want, using the Settings app. A community member partially implemented this: https://github.com/pop-os/cosmic-files/pull/1555 & https://github.com/pop-os/cosmic-settings/pull/1860 It got stuck on a bit of a packaging issue: we need to preserve the current (intended, UX-mandated) behavior on upgrade and as the default for future installations, which means we need COSMIC Files to be the default archival app by default (and it seems the author of those PRs wasn't sure how to accomplish that).

krakotay:
> partially implemented this: #1555 & [pop-os/cosmic-settings#1860](https://github.com/pop-os/cosmic-settings/pull/1860) 

Ok, I need some time to understand the PRs. But before the global patch, that makes cosmic files "default archival app by default", what do you think about simple setting pref? Something like "`Extract to` archive by default", on/off ("on" by default, like in master branch, and "off" like in the my patch).

PR #1934, opened 2026-07-18
Honor a per-folder custom icon from .directory (Icon=) — main view + sidebar

### LINKED ISSUE #1427: [Feature Request] Allow using image files as custom folder icons (via .icons directory)
**Issue/Description:**
COSMIC Files currently does not allow assigning a custom icon to a folder using an image file.  This is a request to add support for the icon for a directory, with the icon files stored in a central `.icons` directory.

**Proposed Behavior:**
When listing folders inside a parent directory (e.g., `Games/`), COSMIC Files should check for a hidden `.icons` directory inside that same parent. If a matching image file exists inside `.icons` with the exact folder name, that image should be used as the folder icon.

**Example:**
Games/
- RDR1/
- RDR2/
- .icons/
--- RDR1.ico
--- RDR2.ico

In this case:
- The folder `RDR1` should use `.icons/RDR1.ico`  
- The folder `RDR2` should use `.icons/RDR2.ico`

This avoids cluttering each folder with hidden images and makes icon management simple.

**Expected Behavior:**
If `.icons/<foldername>.png` (or `.jpg` / `.jpeg`) exists inside the parent directory, COSMIC Files should render that image as the folder icon.

**Other Notes:**
This approach:
- Keeps directories clean
- Scales well for large collections (games, media, projects)
- Requires no extra UI or settings

I have these & more icons I would like to use for my game folders
 
Image

tomashaa:
  - **AI disclosure:** this PR was developed with AI assistance (Claude / Cursor). I have reviewed the diff in full, tested it on real hardware (Pop!_OS / COSMIC, running as a patched local install), and I maintain the change and respond to review myself.

---

Lets an app or user brand a folder — the "this folder looks like *X*" effect
Dropbox/Nextcloud/etc. rely on, and that KDE/Dolphin already supports via the
same file. Today COSMIC Files always shows the generic (or special-dir) folder
icon.

`folder_icon()` and `folder_icon_symbolic()` look for a `.directory` file in
the directory and read its `Icon=` key before falling back to the theme folder
icon, resolved through the existing `desktop_icon_handle()` helper (themed name
*or* absolute path — same handling as `.desktop` icons). Because both functions
read it, one `.directory` file brands the folder in the main view **and** its
sidebar favourite entry.

**Scope:** one private helper + an early return in each function. No new
dependencies, no subprocess. `folder_icon()` is the single point all directory
items (filesystem, gvfs, trash) already route through; the read happens once
per item at construction, where mime guessing already runs.

### Follow-up: non-square image icons in the sidebar

Wide raster `Icon=` paths (common for brand logos, e.g. 1920×1201 PNGs) were
still **horizontally squashed** in sidebar favorites even after using
`ContentFit::Contain` for the nav widget.

**Cause:** the nav/segmented-button layout allocates a square icon slot and
draws into it; a non-square image handle still gets forced into that square.

**Fix (`6d01656`):** for absolute-path raster icons, downscale large assets
(cap longest side at 256), then **pad to a square** with transparent margins
before creating the icon handle. SVG and already-square rasters stay on the
cheap `from_path` path. Combined with sidebar `ContentFit::Contain`, the mark
keeps its aspect ratio in both the main view and the sidebar.

Also in this PR:
- Prefer Cosmic/Pop `*-symbolic` theme icons when resolving bare `Icon=` names
- Re-read favorite sidebar icons when activating a nav entry so `.directory`
  changes show without opening a new window

**Try it:**
```
printf '[Desktop Entry]\nIcon=folder-download\n' > ~/somedir/.directory
```
Reopen the parent → `somedir` shows the download-folder icon. An absolute path
(`Icon=/path/to/wide-logo.png`) works too and should no longer look squashed
in the sidebar.

Built and run on Pop!_OS/COSMIC; the icon renders in both the main view and the
sidebar. Happy to also add the GNOME `gio metadata::custom-icon` source behind
the existing gio/gvfs feature in a follow-up if you'd like GNOME parity.

Closes #1427

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -2650,6 +2650,18 @@ impl Application for App {
     }
 
     fn on_nav_select(&mut self, entity: Entity) -> Task<Self::Message> {
+        // Re-read per-folder `.directory` icons when activating a favorite so the
+        // sidebar picks up icon changes without requiring a new window.
+        if self.nav_model.data::<FavoriteIndex>(entity).is_some()
+            && let Some(Location::Path(path)) = self.nav_model.data::<Location>(entity).cloned()
+            && path.is_dir()
+        {
+            self.nav_model.icon_set(
+                entity,
+                icon::icon(tab::folder_icon_symbolic(&path, 16)).size(16),
+            );
+        }
+
         self.nav_model.activate(entity);
         if let Some(location) = self.nav_model.data::<Location>(entity) {
             let should_open = match location {
--- a/src/app.rs
+++ b/src/app.rs
@@ -1778,14 +1778,16 @@ impl App {
                 };
                 nav_model = nav_model.insert(move |b| {
                     b.text(name.clone())
-                        .icon(
-                            icon::icon(if path.is_dir() {
-                                tab::folder_icon_symbolic(&path, 16)
-                            } else {
-                                icon::from_name("text-x-generic-symbolic").size(16).handle()
-                            })
-                            .size(16),
-                        )
+                        .icon(if path.is_dir() {
+                            tab::folder_nav_icon(&path, 16)
+                        } else {
+                            icon::icon(
+                                icon::from_name("text-x-generic-symbolic")
+                                    .size(16)
+                                    .handle(),
+                            )
+                            .size(16)
+                        })
                         .data(match favorite {
                             Favorite::Network { uri, name, path } => {
                                 Location::Network(uri.clone(), name.clone(), Some(path.to_owned()))
@@ -2656,10 +2658,8 @@ impl Application for App {
             && let Some(Location::Path(path)) = self.nav_model.data::<Location>(entity).cloned()
             && path.is_dir()
         {
-            self.nav_model.icon_set(
-                entity,
-                icon::icon(tab::folder_icon_symbolic(&path, 16)).size(16),
-            );
+            self.nav_model
+                .icon_set(entity, tab::folder_nav_icon(&path, 16));
         }
 
         self.nav_model.activate(entity);
--- a/src/dialog.rs
+++ b/src/dialog.rs
@@ -900,16 +900,16 @@ impl App {
                 };
                 nav_model = nav_model.insert(move |b| {
                     b.text(name.clone())
-                        .icon(
-                            widget::icon::icon(if path.is_dir() {
-                                tab::folder_icon_symbolic(&path, 16)
-                            } else {
+                        .icon(if path.is_dir() {
+                            tab::folder_nav_icon(&path, 16)
+                        } else {
+                            widget::icon::icon(
                                 widget::icon::from_name("text-x-generic-symbolic")
                                     .size(16)
-                                    .handle()
-                            })
-                            .size(16),
-                        )
+                                    .handle(),
+                            )
+                            .size(16)
+                        })
                         .data(Location::Path(path.clone()))
                 });
             }
--- a/src/tab.rs
+++ b/src/tab.rs
@@ -320,6 +320,14 @@ pub fn folder_icon_symbolic(path: &PathBuf, icon_size: u16) -> widget::icon::Han
     .handle()
 }
 
+/// Sidebar/nav folder icon that keeps image aspect ratio (pad, don't squash).
+/// Matches the main view, which uses [`ContentFit::Contain`] for item icons.
+pub fn folder_nav_icon(path: &PathBuf, icon_size: u16) -> widget::icon::Icon {
+    widget::icon::icon(folder_icon_symbolic(path, icon_size))
+        .size(icon_size)
+        .content_fit(ContentFit::Contain)
+}
+
 //TODO: replace with Path::has_trailing_sep when stable
 fn has_trailing_sep(path: &Path) -> bool {
     path.as_os_str()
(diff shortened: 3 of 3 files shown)


tomashaa:
Following the feedback on pop-os/libcosmic#1367: I've filled in the PR template and added an explicit AI-assistance disclosure to the commit message (branch force-pushed; diff unchanged). Apologies for missing both initially. If you'd rather implement this in-house as with the opacity change, feel free to close — the .directory convention itself is the part I care about.

KodeBarista:
Works really well. Thanks for implementing this.

Merging this PR should also close this issue https://github.com/pop-os/cosmic-files/issues/1427

image

jacobgkau:
The XDG specification basically just acknowledges that `.directory` files can exist, but doesn't delve into the contents (or even the exact filename) aside from the more general documentation about possible desktop entry keys, of which `Icon` is one: https://specifications.freedesktop.org/desktop-entry/latest/file-naming.html & https://specifications.freedesktop.org/desktop-entry/latest/recognized-keys.html 

So KDE's approach seems about as close to a standard as there could be for this right now. It sounds like GNOME stores icon selections in `~/.local/share/gvfs-metadata` instead, which I don't think we'd want to rely on, although we use GVFS in some places.

jacobgkau (commented):
If a folder's already in the Favorites menu and then the folder's icon is changed, it doesn't update in the Favorites menu in the existing window, only new windows. For custom icons in general, I don't think that's critical, but wanted to mention it as a limitation of this implementation. Not sure if it'd be worth it to re-read the icon at least on navigation into the folder (activation of the sidebar entry) or something.

It seems like it'd be a problem to not be watching the icon somehow for the Dropbox/Nextcloud/etc use-case you mentioned, since the point of those would presumably be to update live with the sync status?

Screenshot_2026-07-23_15-56-20

Another thing I noticed is that the custom icons showing up aren't sourcing from our icon theme. In the above screenshot, the top folder's icon is set to `audio-volume-high`, which you can also see (with the intended theme) in the icon browser on the left. Do you know why that would be?

tomashaa:
Thanks for the review — addressing each point:

### Icon theme mismatch (`audio-volume-high`)
Good catch. Cosmic (and Pop) often ship status icons only as `*-symbolic`. Looking up the bare name was inheriting a foreign full-color icon via the theme chain, so it didn’t match the Cosmic icon browser.

Pushed a fix in `cf04329`: for themed `Icon=` names (absolute paths unchanged), prefer a Cosmic/Pop-family match, and if the bare name isn’t in that family but `{name}-symbolic` is, use the symbolic one.

### Favorites not updating until new window
Agreed it’s a limitation of reading icons at construction time. Also in `cf04329`: when a favorite nav entry is activated, we re-read `folder_icon_symbolic()` for that path so a changed `.directory` shows up without restarting the window. Still not a live file watcher — just a cheap refresh on activation.

### Live watching for Dropbox/Nextcloud sync status
That’s intentionally out of scope for this PR. Here `.directory` / `Icon=` is for **static folder branding** (the KDE-style “this folder looks like X” case). Live sync-status overlays belong with the emblems RFC discussion in #1935 / draft PR #1936.

Happy to adjust further if you’d rather keep the symbolic preference narrower (e.g. only in `custom_folder_icon` rather than shared `desktop_icon_handle`).

jacobgkau (dismissed):
Those changes seem good. Clicking the item's sidebar entry now updates its icon (navigating away and back into the folder without clicking the sidebar entry does not update the icon in the sidebar, though; it's specifically when it's clicked in the sidebar). I'm seeing the correct icons appear now when using icon names.

When using a path to an image file, I noticed that the image is scaled down with padding to maintain its aspect ratio in the main folder view, but it's squashed to be a square in the sidebar. E.g. the green icon below:

Image

This doesn't seem like a serious problem (as it can be worked around by creating a square image file with the desired contents centered), but it is inconsistent, so it's something else that could be addressed now or later.

<details>

## Checklist

### Basic navigation

### File operations

- [X] Files can be renamed.

- [X] Sorting options work.

### External filesystems

### Integrations

</details>

It will be up to @pop-os/engineering whether the AI-generated code is architecturally sound or not.

tomashaa:
Thanks — addressed the image aspect-ratio inconsistency:

Sidebar favorites were using the default icon `ContentFit::Fill`, which squashes wide images into a square. The main view already uses `Contain` (scale with padding).

Pushed `6bd4f3c`: folder nav/sidebar icons now use `ContentFit::Contain` via a small `folder_nav_icon()` helper (also used for dialog nav and re-read-on-activate). Absolute image paths like the green “TESTING” logo should now pad above/below in the sidebar instead of stretching.

`cargo check -j1` passed locally.

tomashaa:
Follow-up on the image aspect-ratio point:

`ContentFit::Contain` alone (`6bd4f3c`) was not enough in practice for wide raster `Icon=` paths in the sidebar — the nav/segmented-button still draws into a square slot, so non-square brand images could still look squashed.

Pushed `6d01656`: for absolute-path raster icons we downscale large assets and **pad to a square** with transparent margins before creating the handle (SVG / already-square images stay on `from_path`). Together with sidebar `Contain`, wide logos keep their aspect ratio in both the main view and favorites.

Tested locally on Pop!_OS / COSMIC with a wide PNG via `.directory` `Icon=`.

PR #1956, opened 2026-08-05
fix: infinite dialogs with network mount failure

### LINKED ISSUE #1629: Impossible to Cancel Out Of Network Access Dialog
COSMIC Files can get into a permanent infinite dialog-loop when accessing a remote network location using an SSL private key with passphrase. 

**Reproduction**

Set up:
1. Have an SSL public/private key set up for remote SSH/SFTP login that [uses a passphrase](https://stackoverflow.com/questions/3818886/how-do-i-add-a-password-to-an-openssh-private-key-that-was-generated-without-a-p). 
   1. Ensure that you have not yet unlocked the private key in the keyring this session.
2. Have a directory on the remote machine set up as a bookmarked location (via Add to Sidebar menu item) from a prior access

(It may also occur with a fresh login; my use case is as above)

Actions:
1. Open COSMIC Files
2. Click on the saved remote directory bookmark in the sidebar. You should get the Access Prompt dialog asking for the private key passphrase
3. Click cancel to dismiss the Access Prompt dialog.
4. Behold a dialog within the Files window ("Unable to access network drive")
5. Witness also that the Access Dialog hath returned.
6. Click cancel on either dialog
7. Behold that the dialog returns promptly to thine screen, as if 'twere a 1995 virus.
8. Wail unto the skies, for lo, the only exit is to enter the password or destroy the Files process

**Additional Notes**
Clicking other directories in Files is apparently allowed (non-modal), but does not change either dialog's behaviour.   

I have not tested this with a password login access instead of SSH+passphrase key. While I suspect that code pathway is different, it's probably worth checking for the same behaviour.

norepro:
Selecting a broken network mount triggers an infinite loop of dialogs that requires killing the app:

1. Click invalid network mount
2. Try to mount network path
3. Show failure dialog
4. Trigger `NetworkDriveOpenEntityAfterMount`
    * BUG: This fires even if the mount failed or the user canceled the dialog!
6. Goto 2

Fix by having `network_drive(...)` return a Boolean indicating whether the mount actually happened and using that to break the infinite error loop.

Fixes #1629 

Sample broken network path that triggers the infinite dialog when selected:

**~/.config/cosmic/com.system76.CosmicFiles/v1/favorites**

```ron
[
    Home,
    Documents,
    Downloads,
    Network(
        uri: "sftp://myserver/shared",
        name: "",
        path: "/tmp/fake-network/sftp-host=myserver/mnt",
    ),
]
```

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -2678,10 +2678,14 @@ impl Application for App {
                         })
                         && let Some(mounter) = MOUNTERS.get(&key)
                     {
-                        return mounter.network_drive(uri.clone()).map(move |()| {
-                            cosmic::Action::App(Message::NetworkDriveOpenEntityAfterMount {
-                                entity,
-                            })
+                        return mounter.network_drive(uri.clone()).map(move |mounted| {
+                            if mounted {
+                                cosmic::Action::App(Message::NetworkDriveOpenEntityAfterMount {
+                                    entity,
+                                })
+                            } else {
+                                cosmic::action::none()
+                            }
                         });
                     }
 
@@ -3586,7 +3590,7 @@ impl Application for App {
                         Some((*mounter_key, self.network_drive_input.clone()));
                     return mounter
                         .network_drive(self.network_drive_input.clone())
-                        .map(|()| cosmic::action::none());
+                        .map(|_| cosmic::action::none());
                 }
                 log::warn!(
                     "no mounter found for connecting to {:?}",
--- a/src/mounter/gvfs.rs
+++ b/src/mounter/gvfs.rs
@@ -659,7 +659,7 @@ impl Mounter for Gvfs {
         )
     }
 
-    fn network_drive(&self, uri: String) -> Task<()> {
+    fn network_drive(&self, uri: String) -> Task<bool> {
         let command_tx = self.command_tx.clone();
         Task::perform(
             async move {
@@ -668,9 +668,15 @@ impl Mounter for Gvfs {
                 command_tx.send(Cmd::NetworkDrive(uri, res_tx)).unwrap();
                 res_rx.await
             },
-            |x| {
-                if let Err(err) = x {
+            |result| match result {
+                Ok(Ok(())) => true,
+                Ok(Err(err)) => {
                     log::error!("{err:?}");
+                    false
+                }
+                Err(err) => {
+                    log::error!("{err:?}");
+                    false
                 }
             },
         )
--- a/src/mounter/mod.rs
+++ b/src/mounter/mod.rs
@@ -114,7 +114,7 @@ pub trait Mounter: Send + Sync {
     fn items(&self, sizes: IconSizes) -> Option<MounterItems>;
     //TODO: send result
     fn mount(&self, item: MounterItem) -> Task<()>;
-    fn network_drive(&self, uri: String) -> Task<()>;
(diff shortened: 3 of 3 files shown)


jklgrasso (approved):
I'm not doing the full checklist since only network drives are being changed.

I was able to recreate the issue this fixes on the master branch. This PR does fix the infinite dialogs when attempting to mount a favorited network drive. Mounting, browsing, and ejecting a network drive with the `ssh` file transfer protocol works. The network mount timeout dialog for non-existent/inaccessible drives works.

chris-010:
Verified this on Pop!_OS with COSMIC, cosmic-files 1.5.0 (`24e34ea`), ThinkPad T490, SMB share.

I compared two release builds of the same tree: **A** = the stock 1.5.0 package binary (which is exactly `24e34ea`), **B** = that tree with this PR applied. Both were run against an isolated `XDG_CONFIG_HOME` holding two bookmarks, so my real config stayed out of it.

**Reproducer for #1629 without needing an SSH key:** a `Favorite::Network` bookmark whose stored path does not exist and whose URI carries a username that is not in the keyring, so gvfs has to ask for a password.

**A (master):** one click on the bookmark, then 11 × Cancel produced 11 fresh password prompts — no further clicks from me. The dialog is modal, so the window is unusable and the only way out is killing the process. Log for that run:

```
12 x  network drive smb://<user>@<server>/home: result Err(Error { domain: g-io-error-quark, code: 30, message: "..." })
12 x  cancelled connection to "smb://<user>@<server>/home"
 0 x  failed to connect
```

Code 30 is `G_IO_ERROR_FAILED_HANDLED`, so at this error type no error dialog is raised at all — the loop consists purely of the auth prompt coming back.

**B (this PR):** 4 clicks, 4 attempts, each Cancel ends it. The window stays usable.

**Positive path still works:** with the share unmounted, clicking the bookmark mounts and navigates (`result Ok(())` → `connected to ...`). Worth testing explicitly, because clicking a bookmark whose share is *already* mounted never touches your change — `path.try_exists()` is true then and `network_drive()` is not called.

The build is clean; the only warning is the pre-existing unused `style` import in `app.rs`.

### Why it loops, in case it is useful for the changelog

It is a direct recursion rather than something in tab loading:

`on_nav_select(entity)` → `mounter.network_drive(uri)` → `.map(|()| NetworkDriveOpenEntityAfterMount { entity })` **unconditionally** → the handler for that message calls `self.on_nav_select(entity)` again → the path still does not exist → same branch again. Cancelling never removes the reason for the next attempt. This PR cuts the cycle at the unconditional `.map`, which is the right place.

### One adjacent case, not a blocker

`Ok(Err(_)) => false` also covers `AlreadyMounted`. In `Cmd::NetworkDrive` the oneshot is sent `Err(...)` for *every* error, while the parallel `Event::NetworkResult` exempts only `FailedHandled`. Measured on build B: connecting to an already-mounted share raises "Unable to access network drive — Location is already mounted" with Cancel / Retry, and Retry can only fail again.

That is pre-existing and orthogonal to what you fix here. It only seems worth flagging because your change gives `false` a new meaning ("do not navigate"), and "already mounted" is one of the cases that maps to it. Treating `AlreadyMounted` as success in that match arm would clear the dialog and keep the new logic exact. Happy to send that as a separate PR if it is wanted.

jackpot51 (approved):

Issue #1160, opened 2025-08-29
Search field doesn't close when changing directories via breadcrumbs bar


brandonbeaumia:
The search field already closes when changing directories from the sidebar or double-clicking a folder in your search results, but not when changing directories in the breadcrumbs bar (even including typing a new path).

iamkartiknayak:
Working for me 

```
cosmic-files:
  Installed: 0.1.0~1758135135~24.04~bdf524c
```

WatchMkr:
> But...if you click any of the links in the breadcrumbs bar at the top (next to and including the back/forward arrows), the search field remains open and filtering your view. This should behave the same as clicking in the sidebar or opening a new folder: closing the search field.

This is by design though perhaps there should improvements to show what's happening. (The design gets a little complex in the single column view).

The reason is to enable expanding the current search by going down a directory if what you queried is not found in the directory you're in.

chris-010:
Still the same on 1.7.0 (`1.7.0~1788189718~24.04~de73942`), checked today against the unmodified package build: with the search field focused, the first Esc only removes focus and the second closes the search. Same with a file selected.

Where the first press goes: libcosmic's text input handles `Named::Escape` by calling `state.unfocus()` and publishing `on_unfocus` if one is set (`src/widget/text_input/input.rs:2230`, rev `2a73fbc` — the revision this tree pins), and that key handler ends in `shell.capture_event()`. So `App::on_escape`, whose last branch is the one that closes the search, never sees that press. cosmic-files sets `on_unfocus` on none of the three search inputs (toolbar and condensed view in `app.rs`, plus the copy in `dialog.rs`), so nothing reports the unfocus back to the app.

Worth noting for whoever picks this up: `on_unfocus` also fires on Tab and on clicking away, so wiring it straight to "close the search" would close the field in cases where it should stay open. The single-press behaviour asked for above needs that distinction, not just the one line.

PR #1966, opened 2026-08-10
fix: empty search field no longer makes a folder look empty

chris-010:
Clicking the magnifier shows an empty folder before a single character has been typed, and clearing
the field by hand leaves it empty as well. The folder looks like it has no contents rather than like
an unfiltered listing.

### Steps to reproduce

1. Open a folder that has items in it.
2. Press Ctrl+F (or click the magnifier). The item list is empty already.
3. Type a few characters, then delete them again. The list stays empty.
4. Only the X button in the field brings the items back.

Same in the file chooser dialog (for example via "Move to…").

### Cause

`Message::SearchInput` hands the term to `search_set()`, which builds `Location::Search(.., "")` for
an empty one, and `scan_search()` (src/tab.rs) returns immediately for an empty term, so the tab has
no items to show. `Message::SearchClear` is the only caller that leaves search mode.

The four-line variant — treat an empty field like the clear button — was tried first and dropped:
whether the field is drawn is derived from the location too, so the field vanishes as soon as the
last character is deleted.

### Change

`Tab::search_term` now holds the contents of the field, separately from the location; `Some("")`
means "open, not filtering". `Location::Search` is only built for a non-empty term, and an empty
term returns the tab to the location that was being searched, so the folder stays visible.

`change_location()` derives `search_term` from the location it moves to, which keeps both
established behaviours: selecting an entry in the nav bar clears and closes the field (#349), while
the breadcrumbs of a search still widen the search one directory up with the field open (#1160).
Reload and history navigation behave like the breadcrumbs.

`src/dialog.rs` has its own copy of `search_get`/`search_set` for the file chooser; it is changed
the same way, otherwise the dialog would keep the old behaviour.

### Tested

Pop!_OS 24.04, built from this branch and run in place of the packaged binary:

- magnifier / Ctrl+F without typing → folder stays visible, field focused
- type, then clear with backspace → items come back, field stays open and keeps focus
- X button → leaves search, as before
- search, then click a breadcrumb → field stays open with its term, search widens to the parent
- search, then the back arrow → same
- search, then a nav bar entry → field closes, folder unfiltered
- file chooser via "Move to…": same six checks


src/app.rs
```diff
@@ -1621,45 +1619,44 @@ impl App {
...
                     _ => None,
                 },
             };
-            if let Some((location, focus_search)) = location_opt {
+            if let Some(location) = location_opt {
                 tab.change_location(&location, None);
-                title_location_opt = Some((tab.title(), tab.location.clone(), focus_search));
+                title_location_opt = Some((tab.title(), tab.location.clone()));
             }
+            tab.search_term = term_opt;
```
BlakeGardner, line 1643: This assignment runs unconditionally, even when no `SearchLocation` could be derived for the current location, which I think introduces a small regression for unsearchable locations — i.e. anything with no `path_opt()` that isn't Recents/Trash, such as the `network:///` browse root (`Location::Network(_, _, None)`).

Before this change, Ctrl+F / the magnifier was a no-op there: `search_set()` couldn't build a `Location::Search`, nothing changed, and the field never appeared. With this change, `search_term` becomes `Some("")`, so `search_get()` returns `Some`, and the search field is drawn and focused — but typing into it can never produce a `Location::Search` (the `search_location` derivation still yields `None`), so the input silently filters nothing. A dead search box is arguably more confusing than the old no-op.

Could this be gated on the location actually being searchable?

Two related notes:

- `src/dialog.rs` has the same pattern and would need the same gating to keep the two copies in sync (flagged inline there as well).
- `focus_field` is computed from `term_opt.is_some()` at the top of the function, so in this case the app would still emit a focus task for a field that isn't rendered. Harmless, but it could be gated the same way for cleanliness.

Gating inside `search_set()` (rather than in `Message::SearchActivate`) covers all callers, including type-to-search (`TypeToSearch::Recursive`), which calls `search_set_active(Some(term))` directly.

src/dialog.rs
```diff
@@ -787,40 +785,43 @@ impl App {
...
                 _ => None,
             },
         };
-        if let Some((location, focus_search)) = location_opt {
+        let mut location_changed = false;
+        if let Some(location) = location_opt {
             self.tab.change_location(&location, None);
+            location_changed = true;
+        }
+        self.tab.search_term = term_opt;
```
BlakeGardner, line 810: Same pattern here — the file chooser copy would need the same gating (see my comment on `search_set()` in `src/app.rs`), otherwise the dialog shows the same non-functional search field on unsearchable locations.

BlakeGardner:
Tested both behaviors side by side, videos attached.

**Before (COSMIC Files 1.7.0):** Pressing <kbd>Ctrl</kbd>+<kbd>F</kbd> or clicking the search magnifier immediately shows "No results found". Typing populates the results as expected, but erasing the search term brings back the "No results found" screen.


https://github.com/user-attachments/assets/1222d4f1-6881-4873-b47d-8b27a0388416

**With this PR:** Pressing <kbd>Ctrl</kbd>+<kbd>F</kbd> or clicking the magnifier keeps the current folder view until you start typing, then results populate. Erasing the search term returns to the original folder view.


https://github.com/user-attachments/assets/7b72e184-f20b-4bde-9fd2-09f0dc3ece9e

chris-010:
Thanks for testing it independently and posting the videos.

That matches what I measured when I opened this: the "No results found" screen shows whenever the search term is empty — both on entering search and after erasing it — in the main window and the dialog alike.

On ordering, in case it saves anyone work: this branch is still mergeable against master and needs no rebase. #2032 (search in network locations) overlaps it in `app.rs` and `dialog.rs` on `search_set`, and you have already offered to rebase if this lands first — so taking this one first is the cheaper order.

src/app.rs
chris-010: You were right on all three points — thanks for the careful read, and sorry for answering the videos and not this.

`Location::Network(_, _, None)` is the only variant that gets there: `path_opt()` returns `None` for it and it is neither Recents nor Trash, so the `network:///` browse root drew and focused a field that could never build a `Location::Search`.

Fixed in 9aedfcb. The term is only stored where a search can actually run, and the gate sits in `search_set()` as you suggested, so type-to-search and every `SearchInput` are covered as well. The focus task is no longer emitted for a field that isn't rendered. The derivation moved to `Location::search_location_opt()` in `tab.rs`, which removes the second copy in `dialog.rs` — your point about keeping the two in step.

Measured on `network:///` before and after the gate: without it the magnifier opens a field and typing filters nothing; with it the magnifier and Ctrl+F are a no-op there again, as they were before this branch. Searching a path, Recents and Trash, widening a search through the breadcrumbs and erasing the last character are unchanged.

Not part of this change: the magnifier button itself is still drawn on a location that cannot be searched, so it is now a silent no-op again as it was before. Happy to gate `.on_press` on the same helper here or in a separate PR, whichever you prefer.

Unrelated to the change: the red check on this PR is not a build failure. The workflow run from 10 August sat in the approval queue for 30 days and was then failed by GitHub without ever starting a job. This push should trigger a fresh run.

src/dialog.rs
chris-010: Same gating here, through the shared helper — details in the reply on the `app.rs` thread.

jacobgkau (commented):
Before:

https://github.com/user-attachments/assets/c1b5879d-af47-4a18-9079-b55cb28b610d

After:

https://github.com/user-attachments/assets/3273290a-c122-44d5-b35e-5a1171914361

Looks fine to me, but @pop-os/ux can you confirm this is okay?

Issue #657, opened 2024-11-14
[BUG] Can't add smb shares under "network" category


Alfatango12:
**Cosmic-files version:**

cosmic-files 1.0.0.alpha.3-1

**Issue/Bug description:**
Adding a network share don't work. It seas "Impossible to access the network drive, volume doesn't implement mount".

**Steps to reproduce:**
1. Open cosmic-files
2. Click the "network" category
3. Click on "add" in the bottom right corner
4. write the ip address starting with smb://
5. click connect
At this point the error window described above appears.

**Expected behavior:**
It should open the smb network share.

**Other notes:**
It works under other file manages like dolphin

alphabitserial:
Do you have gvfs-smb installed?

dbech:
I similarly cannot connect to a samba share on a fresh install of Pop 24.04 alpha 1 which comes with https://github.com/pop-os/cosmic-files/commits/2fa8e6adc44448bd5ac749302154f8f670e7f381

However I get no error following the above method of adding a drive instead it just sits on connecting forever.

If I instead click on the discovered smb server in the networks section then click the share I want it simply appears empty.

Happy to provide any logs etc needed.

bolbit:
Hi same issue here after an upgrade from 22.04, version is https://github.com/pop-os/cosmic-files/commits/63176a1e2a942c7d2f22999f93fc963e2f8039c3.
Works with dolphin.

ricardomillerjr:
Done already installed the gvfs-smb.... Same issue

geoffritter:
The issue: Adding the root host only, e.g. smb://fqdn.or.ip anonymously and without a share path, it will claim it's 'mounted' but does not show any share names. Attempting again says it's already mounted. I would expect the shares to be shown somewhere so that they could be further connected to.

What works: Explicitly putting the share name smb://fqdn.or.ip/yourshare this will ask you for the credentials and connect. It will then add it to the mount list.

Tested in Fedora 43.

OliverWich:
I ran into the same issue on Arch (CachyOS). Installing `gvfs-smb` as suggested fixed it for me using `smb://<ip>/<share name>`.
Just connecting to the host with `smb://<ip>` just put me back into the "Networks" section of COSMIC files without doing anything.

(Sidenote, even with an SMB share mounted, that "Networks" section in COSMIC files remains empty...)

metabubble:
[...]

PR #1972, opened 2026-08-11
fix: open a network share from the server view

chris-010:
Double clicking a share while browsing `smb://server/` does nothing at all, with no message. This is the "click discovered server, click share, it simply appears empty" report in #657.

### Steps to reproduce

1. Select Networks in the sidebar, enter `smb://user@server/` in the location bar.
2. The shares are listed.
3. Double click one of them. Nothing happens, and the log says `no path for item`.

### Cause

Two things meet here.

gvfs enumerates a share as `gio::FileType::Mountable`, not `Directory`, so `is_dir` is false in `network_scan()`. Activation in `tab.rs` navigates when the item is a directory and otherwise wants a path, so a share takes the branch meant for files.

The item has no usable path or uri to offer either, because the child of a browse root is a backend-internal address that can neither be navigated to nor mounted:

```
$ gio list -a standard::target-uri smb://user@server/
home  0  (mountable)  standard::target-uri=smb://server/home
$ gio list -u smb://user@server/
smb://server/._home
$ gio mount smb://server/home      # mounts
$ gio mount smb://server/._home    # exits 0, mounts nothing
```

The item ends up as `Network("smb://server/._home", "home", None)`, and the click ends in `no path for item`.

### Change

The address that works is `standard::target-uri`, and the enumeration already delivers it — `enumerate_children("*")` includes it, it was simply never read. The item now uses it and takes its path from the resolved file, so a share that is already mounted has its local path right away.

A mountable that resolves to a target is treated as a directory, so activating it navigates instead of giving up. Nothing further is needed to mount it: `Cmd::NetworkScan` already mounts the location it is asked to scan when that location is not mounted yet.

`size_opt` is computed lazily now. `bool::then_some` evaluates its argument eagerly, and a mountable has no `standard::size`, so every share in a listing produced a `g_file_info_get_size: should not be reached` critical on stderr.

### Tested

Pop!_OS 24.04, ThinkPad T490, SMB server with six shares. A release build of this branch compared against the stock 1.5.0 package binary of the same commit, both run against an isolated `XDG_CONFIG_HOME`.

Before, with every share unmounted: double clicking a share does nothing, and the log shows `no path for item Item { name: "home", ... location_opt: Some(Network("smb://server/._home", "home", None)) }`.

After:

- the share list shows folder icons instead of generic file icons
- double click on an unmounted share mounts and opens it, and `no path for item` no longer appears
- back, then double click the same share again while it is mounted, opens it directly
- a subfolder inside the share, and back out again, unchanged
- the mounted share appears in the sidebar and opens from there
- `g_file_info_get_size` criticals per server listing: 6 before, 0 after

Not covered: discovered servers under `network:///`, because that list is empty in my network.

### DIFF
--- a/src/mounter/gvfs.rs
+++ b/src/mounter/gvfs.rs
@@ -128,19 +128,29 @@ fn network_scan(uri: &str, sizes: IconSizes) -> Result<Vec<tab::Item>, String> {
         let name = info.name().to_string_lossy().into_owned();
         let display_name = String::from(info.display_name());
 
-        let uri = String::from(file.child(info.name()).uri());
+        let target_uri = info
+            .attribute_as_string(TARGET_URI_ATTRIBUTE)
+            .map(String::from);
+        let child = match &target_uri {
+            Some(target_uri) => gio::File::for_uri(target_uri),
+            None => file.child(info.name()),
+        };
+        let uri = target_uri
+            .clone()
+            .unwrap_or_else(|| String::from(child.uri()));
 
         //TODO: what is the best way to resolve shortcuts?
-        let location = Location::Network(uri, display_name.clone(), file.child(&name).path());
+        let location = Location::Network(uri, display_name.clone(), child.path());
 
         let metadata = if !force_dir && !info.boolean(gio::FILE_ATTRIBUTE_FILESYSTEM_REMOTE) {
             let mtime = info.attribute_uint64(gio::FILE_ATTRIBUTE_TIME_MODIFIED);
-            let is_dir = matches!(info.file_type(), gio::FileType::Directory);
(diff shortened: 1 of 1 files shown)


chris-010:
Closing this rather than leaving it stale — the branch conflicts and is 30 commits behind, and rebasing it would need a fresh round of testing against real shares that I can't justify for a branch this old.

The reasoning stays available in #657: gvfs reports a share as `gio::FileType::Mountable` rather than `Directory`, so `is_dir` is false in `network_scan()` and activation takes the branch meant for files, which wants a path. If that is still the case once the surrounding code settles, I'll re-file small against current master.

PR #1996, opened 2026-08-20
detect remote filesystems so SFTP browsing is not slow

### LINKED ISSUE #582: [Bug] Slow SFTP connections
**Cosmic-files version:**
`0.1.0~12728395411~22.04~c7511bb`

**Issue/Bug description:**
Connecting to a network SFTP connection takes a very long time, this seems to scale with the number of files within a directory. Moving between directories also exhibits this behavior. File transfer speed is fine.

**Steps to reproduce:**
1. Connect to a SFTP server.
2. Experience long connection time / freeze before it gets added to sidebar.
3. Move around in some directories.
4. Experience slow load times.

**Expected behavior:**
To have similar performance to connecting via the command line or another file explorer.

**Other notes:**
SFTP connection is running on a non-default port (if that makes any difference at all).
I have not tested other file protocols such as samba or normal FTP.
The SFTP connection is fast in another file explorer or on the command line.

Running tcpdump on this port while connecting shows a stark difference in packets received when running with cosmic vs another file viewer (in this case I used the default files app on PopOS 22.04). The default file app uses 322 packets and in much shorter timespan. While cosmic uses 823 over a much longer timespan, and the packets appear to be much larger. I don't know if this helps at all.

kxmpxtxnt:
- [x] I have disclosed use of any AI generated code in my commit messages. (didnt use any)

Fixes #582 

filesystem::remote only comes from query_filesystem_info(), but both scan
paths read it from enumerate_children(), where it is always missing. Every
remote file was treated as local, so each guard that skips expensive work on
remote filesystems did nothing.

Also request only the attributes actually used instead of "*", add is_dir to
ItemMetadata::GvfsPath, and count children in the background for visible items.

### DIFF
--- a/src/mounter/gvfs.rs
+++ b/src/mounter/gvfs.rs
@@ -18,6 +18,15 @@ use crate::tab::{self, ChecksumState, DirSize, ItemMetadata, ItemThumbnail, Loca
 
 const TARGET_URI_ATTRIBUTE: &str = "standard::target-uri";
 
+// Attributes requested when listing a network directory.
+const SCAN_ATTRIBUTES: &str = "standard::name,\
+standard::display-name,\
+standard::type,\
+standard::size,\
+standard::icon,\
+standard::is-hidden,\
+time::modified";
+
 fn resolve_uri(uri: &str) -> (String, gio::File) {
     let file = gio::File::for_uri(uri);
     // Resolve the target-uri if it exists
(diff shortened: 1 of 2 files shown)


jacobgkau (approved):
This results in a very significant speedup when browsing SFTP filesystems. Before:

https://github.com/user-attachments/assets/30562f9e-7162-4581-8271-ffc164ec6d20

After:

https://github.com/user-attachments/assets/0be0891d-ae49-4191-94ed-72b7fbb649f4

Thank you for this improvement!

Manipulating files over the SFTP share also works as expected.

<details>
<summary>Standard regression testing passed:</summary>

## Checklist

### Basic navigation

### File operations

- [X] Files can be renamed.

- [X] Sorting options work.

### External filesystems

### Integrations

</details>

jackpot51 (approved):

chris-010:
A note on the attribute list this PR introduced: `SCAN_ATTRIBUTES` does not contain `standard::target-uri`, and `network_scan()` is the only caller.

Nothing in that function reads the attribute today, so this is not a regression. It matters for the "a share in a server listing cannot be opened" half of #657: gvfs reports a share as `Mountable`, and its enumerated child is the backend-internal `smb://server/._share`, which cannot be navigated to. `standard::target-uri` is what resolves it — `resolve_uri()` already relies on that attribute for a URI typed into the path bar, but the listing path can no longer see it. Any fix for that failure would need one more line in the list.

Issue #1400, opened 2025-11-27
[Bug] Ongoing copy/move operation not shown after closing active window


KodeBarista:
**Issue/Description:**
When a copy or move operation is in progress, the active COSMIC Files window shows the transfer status at the bottom.  
If that window is closed while the operation is still running in the background, opening a new COSMIC Files instance does not display the ongoing operation’s status.

**Version:**
```
 cosmic-files:
  Installed: 0.1.0~1764029329~24.04~e053db3
```

**Steps to Reproduce:**
- Start a copy or move operation in COSMIC Files
- Observe the progress indicator at the bottom of the window
- Close the active Files window while the operation is still running
- Open a new Files window
- Note that the ongoing operation is no longer shown (silently running in bg)

**Expected Behavior:**
Any new instance of COSMIC Files should retain and display the current background copy/move operation’s status until it completes, regardless of which window was originally used to start the transfer.

**Other Notes:**
- Showing a number (`n`) near the search icon, like it did before, would be a better approach to notify users that `n` operations are currently in progress. 
- It would be useful to get a notification after a long operation has been completed.

PR #1998, opened 2026-08-22
fix/1400: Long-running operation progress indicator 

MikhailValerie:
Show an indicator in the window header bar when there is a long-running operation. Hovering over the indicator will show the operation and the progress text, with live updates to the progress percentage and it can be clicked to restore the dismissed progress footer bar.

This solves several UX issues with long-running operations:
- ☑️ Cannot view long-running operation after dismissing footer bar
- ☑️ Cannot cancel / pause / resume long-running operation after dismissing footer bar
- ☑️ Better tracking of long-running operations on windows in the background
- ☑️ Cannot view long-running operation after closing active window
- ❌️ Cannot cancel / pause / resume long-running operation after closing active window

https://github.com/user-attachments/assets/493a09f6-48e0-4ebf-9885-d086e5ae57a0

### Indicator icon
When any window has a running operation, every window (but not the Desktop or Dialogs) will display a spinning `indeterminate` indicator in the accent colour, located in the top-right like the **Cosmic Store**.
image

### Working Tab
If the current tab is currently working one or more operations, the tooltip for the icon will replicate the current progress bar title.  Clicking the icon will restore the dismissed progress footer for that window only.  There can be multiple windows each with their own operation running.
Screenshot_2026-08-27_15-55-25

### Other Tabs
Tabs without any current operation will still show the indicator with a title of "X operations running in the background".
Screenshot_2026-08-27_15-54-50

### State
This shares information between windows by writing to a common State file, which is then subscribed to to track changes.  Exiting the app will reset this count to 0 for next time.

### Performance
There was some changes in handling operations that reduce unnecessary repeat rescanning of tabs.  This is not essential for this PR and have been isolated to commit # `c617e975d49e55ad84a91db70b148ae1604741fa` which can be safely reverted.

### Splitting the Work
The current work is suitable to merge; the global queue and progress indicator need more work (potentially upstream) which I'll place into a separate PR.

### Concerns
These are the concerns identified when testing the current code:
 -  Application crashes will potentially leave the a number in the State file, which will trigger the indicator on next application startup.  Cleanly closing the app will reset the counter for the future.
 - Closing a window with an active operation still makes it impossible to pause / resume / cancel the operations triggered by that window; this will be handled via a different PR
 - The icon only indicates activity and doesn't indicate progress.
 - Clicking the progress icon in a window with no active activities currently does nothing.

Relates to: https://github.com/pop-os/cosmic-files/issues/1400

---

- [ ☑️ ] I have disclosed use of any AI generated code in my commit messages.
- [ ☑️ ] I understand these changes in full and will be able to respond to review comments.
- [ ☑️ ] My change is accurately described in the commit message.
- [ ☑️ ] My contribution is tested and working as described.
- [ ☑️ ] I have read the [Developer Certificate of Origin](https://developercertificate.org/) and certify my contribution under its conditions.


KodeBarista:
Would it not be better to use a deterministic progressbar which reflects the current progress? Similar as nautilus.

MikhailValerie:
> Would it not be better to use a deterministic progressbar which reflects the current progress? Similar as nautilus.

That's a great point.  The progress indicator currently:
- Indicates there is a long-running operation in progress
- Indicates the progress via mouse-hover tooltip
- Reveals the progress by clicking to show the footer bar

But a deterministic progressbar would, in addition to the above:
- Reveal the current progress without other action when the footer bar is hidden
- Improve watching long-standing operations in the background via the header

I did not implement this because I was keeping this feature consistent with the **cosmic-store** indicator.  They both use the same standard indicator widget and a new widget would need to be created for the deterministic progress bar (unless you can point me to one). 

If a widget doesn't already exist, I believe that it should not be implemented in this PR, but rather a separate issue should be raised against **cosmic-epoch** for a deterministic indicator widget and then apply it system-wide wherever relevant.

MikhailValerie:
Rebased on `epoch-1.8.0`

PR #1993, opened 2026-08-20
Item text and background colouring

MikhailValerie:
Improved the item background and text colouring to be cleaner, smarter, more legible and to work better with Frosted Glass.

In refactoring the Grid View item structure and colouring it looks like I have also fixed several other issues ( listed below ). I'm happy for those fixes to be pulled out individually if desired.

### Text Legibility ( https://github.com/pop-os/cosmic-files/issues/1989 )
Item label text now checks for legibility and switches colour if necessary, both in list view and Grid View.  This is far more noticeable when using extreme custom accent colours like ( #000000 ).

_screenshot: this branch (top) compared to original 1.5.0 (old)_
image

### Grid View Highlights
Highlighted and selected items in Grid View now use the accent colour for the background, with partial transparency for those using Frosted Glass.
image

### Hidden Items ( https://github.com/pop-os/cosmic-files/issues/299 )
Hidden items now have a mild opacity on the text and icon colour (the latter not applied) to have a visual difference.  This is very mild for readability concerns, but it's an easily changed `alpha_scale()` value now.
image

### Cut Items
Items that have been cut now have additional mild opacity applied and show a mild outline to make them stand out more.
image

### List View corners
The corners of List View items now respect the user's Style choice, matching other elements like the sidebar items. ( see screenshots above )

### Frosted Glass ( https://github.com/pop-os/cosmic-files/issues/1914 )
The highlighting for both Grid View and List View is now partially transparent, which also propagates into the Search Results.  This allows those who like less frosting (not me) to see through the list results like other UI elements like the sidebar.
image

### Grid View Drag to Select ( https://github.com/pop-os/cosmic-files/issues/1808 )
The effective mouse areas for Grid View items has been updated:
 - Drag to move will now only happen within the coloured background area
 - The Drop Target for each item is still the larger area to handle the preferred action
 - The Drag to select can now occur in the 'empty' areas between the visible list items (that are actually inside the grid cell), giving more room for that action
image
image

### Icons
Although the icon colouring has been updated in code, the visible icons will remain the same until SVG Icons and Icon::opacity come in future updates ( see: https://github.com/pop-os/cosmic-files/pull/1936 )

Relates to: 
- https://github.com/pop-os/cosmic-files/issues/1989
- https://github.com/pop-os/cosmic-files/issues/299
- https://github.com/pop-os/cosmic-files/issues/1808
- https://github.com/pop-os/cosmic-files/issues/1914

----

[...]

Issue #555, opened 2024-10-03
[Feature Request] Compact view support


bobcat141:
Compact view is a view mode that displays files on a 2D grid that goes top to bottom, left to right taking only enough height to display the file name.

It's useful for showing more files than both icon view and list view while also focusing more on making the filename visible than showing the thumbnail. It's perfect for situations where the filename is more relevant than the thumbnail.

Examples:

![image](https://github.com/user-attachments/assets/cfae9eca-1a64-47f0-92ab-f10789c9a834)

![screenshot](https://github.com/user-attachments/assets/c7c135c3-751c-4822-a5f9-8582947ed4ce)

iamkartiknayak:
Compact view is already implemented in `Settings > Desktop > Appearance`

![image](https://github.com/user-attachments/assets/8a06d992-be92-409e-a0e1-295309806152)

git-f0x:
They're not referring to Interface density, but the layout of files/folders (different than Grid or List).

iamkartiknayak:
Doesn't it change with thd setting?

HuckleHubbing:
Thunar is still my default file manager, for only two reasons. The compact view and bulk file rename ability. I keep trying to use list or icon view but it's just so inefficient once you've experienced compact view. A view like this gets my vote to get me to come over to cosmic-files.

MikhailValerie:
I have created a draft PR ( https://github.com/pop-os/cosmic-files/pull/2005 ) for adding Column View to COSMIC Files.  I'd love to get your feedback on the implementation while I'm polishing up the PR.

PR #2005, opened 2026-08-25
feat: Column View

MikhailValerie:
Enabled Column View as a third view option to List View and Grid View, providing a traditional sideways-scrolling compact view of items.  Items are sorted column-then-row which is different from the standard row-then-column sorting, but goes with the alternate scrolling direction.
image

### Switching View
The "View" menu now shows three view options sorted correctly.  The keyboard shortcut for Column View is **[Ctrl] + [3]**.
image

### More Items
This view is much more compact, allowing one to view 1,000% more items than the list view and about 25% more than grid view on a Full HD monitor when comparing the "compact" spacing.  The Column View is much more readable at these scales than the Grid View.

image

### Item Names
The item widths have been set to be wide enough to show most common file names without issues and still be legible when applying ellipsis to the middle of long filenames.  More items an fit in with a shorter name, but then they become harder to read.

### Dependencies
This is reliant on https://github.com/pop-os/cosmic-files/pull/1993 being pulled first.

### Reasons for Draft
There are a few reasons why this is not ready to merge yet:
1. I need to rebase this PR (and the dependencies) on the latest master - there are some i18n conflicts
2. Scrolling vertically does not move the horizontal scrollbar, and I think it should;
3. Performance issues - my testing folder of 15,400 items is significantly more performant than the original views despite showing more items, but I'd like to improve things further - more visible items has a bigger impact;
4. This is a big PR - I'd like to break it (and the dependency PR) into smaller chunks easier to review; and
5. To solicit feedback on the concept and direction before finalisation.

Relates to: https://github.com/pop-os/cosmic-files/issues/555

----

- [ ☑️ ] I have disclosed use of any AI generated code in my commit messages.
- [ ☑️ ] I understand these changes in full and will be able to respond to review comments.
- [ ☑️ ] My change is accurately described in the commit message.
- [ ☑️ ] My contribution is tested and working as described.
- [ ☑️ ] I have read the [Developer Certificate of Origin](https://developercertificate.org/) and certify my contribution under its conditions.


HuckleHubbing:
Looks great so far! Thank you for putting this together.

The only additional suggestion I have on this is related the width. Looks like it truncates, via ellipses, anything longer than about 24 characters. My personal preference is to show the full name, regardless of length—or at least have an option to. 

I forget if this shortening is common behavior in other file managers that have it, like Nemo, but in Thunar the column is extended to the length of the longest name. Whereas the other columns stay as narrow as the names in the other columns allow. I suspect the dynamic column widths is tougher. 

In order of preference: 

1. Dynamic column widths to show the full name, regardless of length. 
2. The way you've drafted it with a set column width and truncating names when needed. 
3. All the columns expand to the longest name. This latter option was done by one file manager I used briefly and I hated that. 

Of course, if time and effort are of no concern and freely available, having an option in the settings for the 1st or 2nd choice would be ideal.

bobcat141:
I think it looks really neat!

I'd propose a middle ground between the OP and @HuckleHubbing and suggest the width of a grid item to be able to expand to accommodate a longer file name but be limited to the size of the parent view.

Dolphin, for example, lets the item be larger than the width of the parent view and it doesn't make for the best experience:
- You have to scroll a lot more in-between columns with the mouse.
- Using the arrow keys to navigate can also be a bit jumpy.

Both behaviors can be seen in the following video:

<details>

<summary>Recording of Dolphin when filenames don't fit the window</summary>

https://github.com/user-attachments/assets/7f9914b1-c5f8-4fca-aca6-c53150657be3

</details>

But regardless of how this ends up being implemented, I think it will be a great addition to the file manager!

PR #2006, opened 2026-08-25
fix: retry thumbnails after file changes and stale fail markers

### LINKED ISSUE #1997: [BUG] Stale video thumbnail failure markers prevent later retries
**Cosmic-files version:**

```text
Installed: 1.5.0~1786455995~24.04~5bdfffa
Apt candidate inspected: 1.6.0~1787090492~24.04~3c9a78a
```

The installed package was used for the behavioral tests. The relevant negative
cache logic remains present in the inspected `1.6.0` candidate source.

**Issue/Bug description:**

When an external thumbnail command fails, COSMIC Files writes a versioned
failure marker. For an unchanged source URI, modification time, and size, that
marker prevents later thumbnail attempts in the same COSMIC Files version.

This can leave a valid video showing the generic icon after a transient
thumbnailer failure, even when the exact thumbnail command now succeeds.
Reopening the folder does not recover it. Moving only the matching failure
marker aside and reopening the folder causes COSMIC Files to generate and show
the thumbnail normally.

The freedesktop Thumbnail Managing Standard intentionally defines
application-and-version-specific failure markers to avoid repeatedly processing
known unsupported, broken, or prohibitively expensive files. This report does
not propose removing that mechanism. The issue is that COSMIC Files gives any
unsuccessful external process attempt the same persistent treatment, including
a temporary failure that succeeds on a later identical invocation.

No private media, filenames, tag values, or local identifiers are included
with this report. Failure-marker PNGs should not be attached because their
metadata contains the source URI.

**Steps to reproduce:**

This is a recovery/retry reproduction for an existing failure marker. The
original transient process failure was not captured, so this report does not
claim a deterministic trigger for that first failure.

1. Find a readable video that shows a generic icon and has a current-version
   failure marker but no successful thumbnail cache entry.
2. Confirm that the same thumbnailer entry point and registered size now exit
   successfully. Direct output to `/dev/null` to avoid retaining a private
   probe frame:

   ```bash
   FILE=/path/to/affected-video.mp4
   cosmic-player --thumbnail /dev/null --size 512 "$FILE"
   echo "$?"
   ```

3. Reopen the directory in COSMIC Files. The generic icon remains.
4. Calculate and inspect the matching negative-cache entry:

   ```bash
   uri=$(gio info "$FILE" | sed -n 's/^uri: //p' | head -n1)
   key=$(printf '%s' "$uri" | md5sum | cut -d' ' -f1)
   marker="$HOME/.cache/thumbnails/fail/cosmic-files-1.5.0/$key.png"
   test -f "$marker" && echo 'failure marker present'
   ```

5. Back up only that marker outside the active thumbnail cache, then reopen the
   directory.
6. COSMIC Files now generates a valid thumbnail and displays it in place of the
   generic icon.

**Expected behavior:**

A transient external-thumbnailer failure should be retried after a bounded
interval, on an explicit refresh, or in a later COSMIC Files session. A valid
unchanged file should not remain suppressed for the lifetime of that version's
failure marker.

Persistent negative caching can remain useful for deterministic unsupported or
invalid files, provided it does not make temporary process or environment
failures effectively permanent.

**Other notes:**

### Redacted local evidence

- The inspected folder contained both successful video thumbnails and generic
  video icons.
- A filename-free sweep exercised the exact registered thumbnail command
  against all 74 current-version failure markers in the folder, including one
  source that also had a successful cache entry. Thumbnail output was directed
  to `/dev/null`, so no private frames were retained.
- 67 of 74 commands exited `0`. Only seven remained repeatable failures: four
  invalid-framerate cases, two structurally empty or near-zero-duration inputs
  reported generically as missing-plugin failures, and one 1 ms extreme-rate
  input that timed out.
- For one affected H.264/AAC MP4, the exact `--size 512` command exited `0`.
  After only its 1x1 failure marker was moved to a backup and the folder was
  reloaded, COSMIC Files produced a valid 512-pixel PNG and the thumbnail became
  visible.
- The source media was not modified. The marker remains preserved locally for
  rollback and inspection.

### Media-metadata classification

A filename-free metadata matrix compared cache state with container, codec,
profile, pixel format, resolution, stream counts, duration, start time,
framerate fields, permissions, and metadata tag *keys* (never values).

Common successful and failed entries substantially overlapped: H.264/AAC,
HEVC/AAC, VP9/Opus, MP4 and WebM containers, common resolutions, ordinary
one-video/one-audio layouts, file modes, and common tag-key sets all appeared
on both sides. No ordinary metadata field separated the 67 successful retries
from existing failure markers.

The four repeatable invalid-framerate cases were different. GStreamer reported
`Frame rate: 0/1` for each, while FFmpeg reported a nonzero average rate and
nonzero packet durations. The video library used by COSMIC Player rejects a
zero framerate after reading the negotiated appsink caps. COSMIC Player then
maps that underlying open error to the generic text `missing required plugin`
in thumbnail mode:

- <https://github.com/wash2/iced_video_player/blob/cb3635192f3dec37619855c904a168ca37a3dac3/src/video.rs#L272-L301>
- <https://github.com/pop-os/cosmic-player/blob/23d59445afaf74aa0ac39a16d6d88f5f20d91d1f/src/thumbnail.rs#L18-L21>

This confirms that some media can fail deterministically because of negotiated
stream properties. It does not account for the much larger class of unchanged
files whose identical command now succeeds. Those stale markers require a
separate retry/revalidation path even if deterministic media failures continue
to be cached.

### Privacy-safe controls

Locally generated H.264/AAC MP4 and VP9/Opus WebM fixtures both generated
thumbnails successfully. Five rounds of four concurrent commands against those
fixtures also passed 20 of 20 attempts. The installed source limits thumbnail
generation to four workers. Concurrency at that limit or codec choice alone was
therefore not sufficient to reproduce the original transient failures, and
this report does not assign that initial cause:

- <https://github.com/pop-os/cosmic-files/blob/5bdfffa494e6f5ceaa18bc8940d5006bd686d4b4/src/tab.rs#L85-L88>

### Source-level path

The installed source:

1. Returns `CachedThumbnail::Failed` whenever a valid failure marker exists:
   <https://github.com/pop-os/cosmic-files/blob/5bdfffa494e6f5ceaa18bc8940d5006bd686d4b4/src/thumbnail_cacher.rs#L62-L88>
2. Validates both successful thumbnails and failure markers using source URI,
   modification time, and size, without an age or retry condition:
   <https://github.com/pop-os/cosmic-files/blob/5bdfffa494e6f5ceaa18bc8940d5006bd686d4b4/src/thumbnail_cacher.rs#L202-L288>
3. Creates a failure marker whenever thumbnail generation returns no image for
   a MIME type with a registered thumbnailer:
   <https://github.com/pop-os/cosmic-files/blob/5bdfffa494e6f5ceaa18bc8940d5006bd686d4b4/src/tab.rs#L2189-L2203>
4. Treats any nonzero external-thumbnailer status as a failed generation:
   <https://github.com/pop-os/cosmic-files/blob/5bdfffa494e6f5ceaa18bc8940d5006bd686d4b4/src/tab.rs#L2243-L2293>

The failure-marker mechanism comes from the freedesktop
[Thumbnail Managing Standard](https://specifications.freedesktop.org/thumbnail/latest/failures.html),
which describes preserving failures to avoid repeatedly retrying files that are
unknown, broken, or too expensive to thumbnail. It does not define how an
application should distinguish a durable content failure from a temporary
external-process or environment failure.

The inspected `1.6.0` candidate retains the same behavior:

- <https://github.com/pop-os/cosmic-files/blob/3c9a78ac9f81917c53db6273e491ca5206a5c8f0/src/thumbnail_cacher.rs#L74-L88>
- <https://github.com/pop-os/cosmic-files/blob/3c9a78ac9f81917c53db6273e491ca5206a5c8f0/src/tab.rs#L2189-L2203>

The same failure-marker path is also present in upstream `master` at
`28546795b0f4ff65d0c96d98a503b1dfec0d8e8e` (2026-08-20), inspected but not
behaviorally tested:

- <https://github.com/pop-os/cosmic-files/blob/28546795b0f4ff65d0c96d98a503b1dfec0d8e8e/src/thumbnail_cacher.rs#L75-L84>
- <https://github.com/pop-os/cosmic-files/blob/28546795b0f4ff65d0c96d98a503b1dfec0d8e8e/src/tab.rs#L2192-L2200>

### Possible fix directions

- Keep the first external-process failure session-local, and persist a failure
  marker only after a bounded repeated failure.
- Give failure markers a bounded retry age while retaining URI/mtime/size
  validation.
- Let an explicit refresh retry negative-cache entries.
- Distinguish deterministic decode/unsupported-media failures from temporary
  process, resource, or environment failures before writing a persistent
  marker.

Related reports such as
[#501](https://github.com/pop-os/cosmic-files/issues/501) and
[#1950](https://github.com/pop-os/cosmic-files/issues/1950) concern thumbnail
resource usage and process concurrency. The installed four-worker limit and
the 20-of-20 bounded concurrency control distinguish this report from those
issues. Neither report covers successful retries being suppressed by the
negative cache.

CarlosZambonii:

AI disclosure: this fix was developed with the assistance of Claude Code (an AI coding assistant). I reviewed the generated diff line by line, understand the mechanism it relies on, and personally ran the manual verification described below on a real build before opening this PR.

Observed behavior

When cosmic-files fails to generate a thumbnail for a video, it never retries that file, even after the root cause is gone (missing codec installed later, file replaced, permissions fixed). The fail marker is effectively permanent.

Root cause

There were two independent bugs.

1. In-memory state never reset (src/app.rs:3735). The filesystem watcher handler for Message::NotifyEvents already refreshed an item's on-disk metadata when the file changed, but left a TODO instead of clearing item.thumbnail_opt. The thumbnail generation loop skips any item where thumbnail_opt.is_some() (src/tab.rs:7065-7067), so once an item was marked NotImage after a failure, it stayed that way for the rest of the session, even though the on-disk fail marker had already become invalid through its own mtime/size check (src/thumbnail_cacher.rs:202-289). The cache was correct, the UI just never asked it again.

2. On-disk fail marker has no expiry for failures unrelated to the file itself. The fail marker follows the freedesktop Thumbnail Managing Standard and stores Thumb::URI, Thumb::MTime and Thumb::Size, invalidating only when the source file changes. A failure caused by something external to the file, such as a missing thumbnailer that gets installed later, never changes the video file, so the marker never expired.

Approach chosen

1. src/app.rs: replace the TODO with item.thumbnail_opt = None when the watcher detects Data or Metadata changes on a file. This alone fixes the case where the file itself changes while the folder is open, with no tradeoff: it just makes the in-memory state match what the disk cache already knew.

2. src/thumbnail_cacher.rs: add a 24 hour TTL on the fail marker (FAIL_MARKER_TTL). is_fail_marker_valid() checks the marker file's own mtime; past the TTL, or if the marker's mtime is in the future (clock skew, a restored backup), it is treated as expired and the thumbnail is retried, regardless of whether the source file changed. This covers failures whose cause was external to the file. The TTL value is intentionally hardcoded and not user configurable for now, to keep the fix minimal; it can be exposed later if there is demand for a different value.

Each retry that fails calls create_fail_marker() again, which rewrites the marker file, so its mtime resets and the 24 hour window restarts. A file that is genuinely and persistently broken is therefore retried at most once per day, not in a tight loop.

These two changes are independent. Item 1 fixes the bug described in the issue title with no downside. Item 2 is the part that involves a judgment call (the TTL value) and can be reviewed or adjusted separately without affecting item 1.

Alternative considered

Relying only on the existing freedesktop mtime/size invalidation, with no TTL. This was rejected because the most common cause of a video thumbnail failure, such as a missing or crashing external thumbnailer, does not change the video file at all. A marker invalidated only by mtime/size would never expire in that case, which is the exact behavior reported in the issue.

Manual verification

To force a thumbnail failure, shadow the thumbnailer's own binary through PATH, rather than editing any .thumbnailer file. Thumbnailer discovery in src/thumbnailer.rs:75-149 collects entries from every XDG data directory into one list per mimetype and tries each in order, with no deduplication by filename or app id. A broken entry placed in $XDG_DATA_HOME/thumbnailers therefore does not shadow the real one in /usr/share/thumbnailers: both get collected, and the working system entry still succeeds after the broken one fails. Overriding XDG_DATA_DIRS entirely to hide the real entry also does not work cleanly: the xdg-mime crate used for mime type detection in this app (src/mime_icon.rs) reads the same XDG_DATA_DIRS and XDG_DATA_HOME variables to locate the shared MIME database, so hiding the thumbnailer this way risks breaking mime detection for the file being tested too. Replacing the thumbnailer's Exec binary on PATH avoids all of this: command resolution by PATH always takes the first match, so it reliably fails, and it can be flipped back to the real binary without restarting anything.

Reproduction A, in-memory state bug:
1. Find the Exec binary of the relevant thumbnailer, for example grep Exec /usr/share/thumbnailers/*.thumbnailer. On this system it is cosmic-player.
2. Create a fake binary ahead of it on PATH:
   mkdir -p /tmp/fakebin
   printf '#!/bin/sh\nexit 1\n' > /tmp/fakebin/cosmic-player
   chmod +x /tmp/fakebin/cosmic-player
3. Launch cosmic-files with that PATH override and open a folder containing a video: PATH="/tmp/fakebin:$PATH" cosmic-files /path/to/folder
   The item shows a generic icon and a fail marker appears under `~/.cache/thumbnails/fail/cosmic-files-<version>/`.
4. Without closing the folder or restarting the app, make the fake binary pass through to the real one: printf '#!/bin/sh\nexec /usr/bin/cosmic-player "$@"\n' > /tmp/fakebin/cosmic-player
   Then overwrite the video file in place with different content (for example cat other_video.mp4 > video.mp4, not mv, so the file keeps the same inode and produces a Modify event instead of a Create/Delete pair).
5. Before this fix: the item keeps showing the generic icon forever, since the only branch that resets thumbnail_opt was a bare TODO. After this fix: within a couple of seconds the real thumbnail appears, without reopening the folder. The debug log (RUST_LOG=cosmic_files=debug) shows a Modify(Data(Any)) event with no accompanying rescan_tab, confirming the fix's own code path handled it rather than a full tab reload.

Reproduction B, TTL for environmental failures:
1. Repeat steps 1 to 3 above to get a fail marker on disk.
2. Restore the real binary on PATH by running only the printf command from step 4 above (the one that makes the fake cosmic-player exec into the real binary). Do not run the video overwrite from that same step 4, and do not touch the video file in any way: this reproduction is specifically testing that the marker expires by age alone, not by a file change.
3. Before this fix: closing and reopening the folder, or restarting the app, still shows the generic icon forever, since the fail marker's Thumb::MTime and Thumb::Size still match the unchanged file.
4. After this fix: age the marker past the TTL without touching the video:
   `touch -d "-25 hours" ~/.cache/thumbnails/fail/cosmic-files-<version>/<hash>.png`
   Reopen the folder or restart the app. The thumbnail is retried and appears correctly, even though the video file's mtime and size never changed.

Both reproductions above were run against a real build of this branch, verified through the debug log and the cache directory contents rather than only by inspection.

Automated tests

cosmic-files has no integration tests that exercise a real thumbnailer or open the UI, so this fix adds unit tests for the pure TTL comparison logic (is_fail_marker_expired), which is deterministic and does not depend on the system clock or filesystem state. The rest is covered by the manual reproduction above.

cargo build --release, cargo fmt and cargo clippy all pass with no new warnings.

Fixes #1997

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -3732,7 +3732,10 @@ impl Application for App {
                                                             );
                                                         }
                                                     }
-                                                    //TODO item.thumbnail_opt =
+                                                    // Force the thumbnail to be regenerated, since the
+                                                    // file's data or metadata (mtime/size) changed and any
+                                                    // previous thumbnail or failure marker may be stale.
+                                                    item.thumbnail_opt = None;
                                                 }
                                             }
                                         }
--- a/src/thumbnail_cacher.rs
+++ b/src/thumbnail_cacher.rs
@@ -8,9 +8,16 @@ use std::io::{self, BufReader, BufWriter};
 use std::os::unix::fs::PermissionsExt;
 use std::path::{Path, PathBuf};
 use std::sync::LazyLock;
-use std::time::UNIX_EPOCH;
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
 use tempfile::NamedTempFile;
 use url::Url;
+
+/// Fail markers older than this are ignored, so a file that previously failed
+/// to thumbnail (e.g. because a thumbnailer was missing) is retried even when
+/// the source file itself never changed.
+// 24h: long enough to avoid hammering a file that keeps failing, short enough
+// that a fix (e.g. installing a thumbnailer) is picked up the same day.
+const FAIL_MARKER_TTL: Duration = Duration::from_secs(24 * 60 * 60);
 
 /// Implements thumbnail caching based on the freedesktop.org Thumbnail Managing Standard.
 /// <https://specifications.freedesktop.org/thumbnail-spec/latest>/
@@ -80,7 +87,7 @@ impl ThumbnailCacher {
         }
 
         // Check if there is a fail marker from an earlier failure.
-        if self.is_thumbnail_valid(&self.thumbnail_fail_marker_path) {
+        if self.is_fail_marker_valid() {
             return CachedThumbnail::Failed;
         }
 
@@ -287,6 +294,39 @@ impl ThumbnailCacher {
 
         true
     }
+
+    /// Like `is_thumbnail_valid`, but also treats the fail marker as invalid
+    /// once it's older than `FAIL_MARKER_TTL`. This covers failures caused by
+    /// a transient or environmental issue (e.g. a missing thumbnailer that
+    /// gets installed later), where the source file's mtime/size never
+    /// change and so would otherwise never be retried.
+    fn is_fail_marker_valid(&self) -> bool {
+        match fs::metadata(&self.thumbnail_fail_marker_path).and_then(|m| m.modified()) {
+            Ok(marker_mtime) => match SystemTime::now().duration_since(marker_mtime) {
+                Ok(age) if is_fail_marker_expired(age) => return false,
+                // Age within the TTL: fall through to the normal validity check.
+                Ok(_) => {}
+                Err(_) => {
+                    // The marker's mtime is in the future (clock skew, a
+                    // restored backup, etc). Don't trust a clock that's
+                    // wrong: treat the marker as expired so we retry
+                    // instead of getting stuck on it indefinitely.
+                    return false;
+                }
+            },
+            Err(_) => {
+                // No fail marker on disk, or its metadata can't be read.
+                // `is_thumbnail_valid` below will correctly return false
+                // in that case too, causing a retry.
+            }
+        }
+
+        self.is_thumbnail_valid(&self.thumbnail_fail_marker_path)
+    }
+}
+
+fn is_fail_marker_expired(age: Duration) -> bool {
+    age > FAIL_MARKER_TTL
 }
 
 fn thumbnail_uri(path: &Path) -> io::Result<String> {
@@ -367,3 +407,22 @@ static THUMBNAIL_CACHE_BASE_DIR: LazyLock<Option<PathBuf>> = LazyLock::new(|| {
 
     None
 });
+
+#[cfg(test)]
+mod tests {
+    use super::{FAIL_MARKER_TTL, is_fail_marker_expired};
+    use std::time::Duration;
+
+    #[test]
+    fn fail_marker_not_expired_within_ttl() {
+        assert!(!is_fail_marker_expired(Duration::ZERO));
+        assert!(!is_fail_marker_expired(FAIL_MARKER_TTL));
+    }
+
+    #[test]
+    fn fail_marker_expired_past_ttl() {
+        assert!(is_fail_marker_expired(
+            FAIL_MARKER_TTL + Duration::from_secs(1)
+        ));
+    }
+}


CarlosZambonii:
Updating this one too for consistency: restored the PR template with all items checked, and added an explicit AI-disclosure line to the commit message (Claude Code was used to help implement and verify this fix; I reviewed and manually tested it myself before opening the PR). Force-pushed the amended commit.

PR #2008, opened 2026-08-25
fix: allow arrow keys to switch focus in delete confirmation dialogs

### LINKED ISSUE #1990: [Bug] It's not possible to switch focus between the buttons on the delete dialog with the arrow keys
**Describe the bug**

When a dialog box is opened to permanently delete a file, it's not possible to navigate from a focused button to a nearby unfocused button with the arrow keys, only with the tab key.

When navigating the files using the arrow keys with the dialog open, the scroll also seems to be defective, as the previously hidden files remain invisible, as can be seen in the evidence video below.

**To Reproduce**

1. Open COSMIC Files.
2. Press shift+delete to permanently delete a file.
3. Press left to focus "cancel".
4. Notice this will change the focused file instead.

**Expected behavior**

It should be possible to change the focus in-between the buttons using the arrow keys.

**Evidence**
<details>

<summary>Video evidence</summary>

Behavior on COSMIC Files:

https://github.com/user-attachments/assets/f8106e13-1b8a-4a89-8c2c-7e98327de28e

Behavior on Thunar, for comparison:

https://github.com/user-attachments/assets/d8ca19e5-a70f-447e-8925-104b66c9c883

</details>

**Desktop (please complete the following information):**

 - OS: Pop!_OS 24.04
 - Package version: `cosmic-files/noble,now 1.5.0~1786455995~24.04~5bdfffa`

**EDIT:** after posting, I've realized the same issue happens to the shut down/restart system dialogs, so maybe it's a libcosmic issue.

CarlosZambonii:

AI disclosure: this fix was developed with the assistance of Claude Code (an AI coding assistant). I reviewed the generated diff line by line, understand the mechanism it relies on (iced/libcosmic's `focus_next`/`focus_previous` operations), and personally ran the manual verification described below on a real build before opening this PR.

Observed behavior

In the "Permanently delete?" and "Empty trash item" confirmation dialogs, Tab/Shift+Tab correctly cycle keyboard focus between the Cancel and Delete buttons, but the arrow keys do nothing, forcing users who navigate by keyboard to rely on Tab alone.

Root cause

Tab-based focus cycling is not implemented by cosmic-files at all: it comes from libcosmic's own app-wide `keyboard_nav` subscription (src/keyboard_nav.rs in the libcosmic crate), which listens for raw Tab/Shift+Tab presses and dispatches `iced::widget::operation::focus_next()`/`focus_previous()`. cosmic-files never wired up an equivalent for arrow keys.

Separately, cosmic-files already binds bare arrow keys to file-list navigation (`ItemLeft`/`ItemRight`/`ItemUp`/`ItemDown` in src/key_bind.rs), dispatched from the app's own `Message::Key` handler in src/app.rs. That handler ran its key-bind loop unconditionally, with no check for whether a modal confirmation dialog was currently on top, so pressing an arrow while the delete dialog was open moved the file grid's selection behind the dialog instead of doing anything to the dialog itself.

Approach chosen

In the `Message::Key` handler, before the file-list key-bind loop runs, check whether the front dialog page is `DialogPage::PermanentlyDelete` or `DialogPage::DeleteTrash`. If so, Left/Up and Right/Down are intercepted and mapped to `operation::focus_previous()`/`operation::focus_next()` and the file-list key binds are never reached while a dialog is open.

`focus_previous()`/`focus_previous()` were chosen over manually focusing a specific button by `widget::Id` (e.g. `cosmic::widget::button::focus(id)`, already used elsewhere in this file to pre-focus the Delete button when a dialog opens) because that's the exact operation libcosmic's own Tab handling already dispatches for this dialog. An earlier version of this fix targeted the Cancel/Delete button ids directly: it worked functionally (confirmed by pressing Enter afterward, which correctly triggered whichever action was focused) but the focus ring never visually updated after the id-targeted `focus()` call, while the pre-existing Tab path (also `focus_next`/`focus_previous`) redraws correctly. Reusing the same operation as Tab sidesteps that redraw discrepancy instead of chasing it further, and also means only one dialog page needs a new helper (`dialog_supports_arrow_focus`) rather than a per-dialog pair of button ids.

Alternative considered

Fixing this upstream in libcosmic's `keyboard_nav.rs`, so every COSMIC dialog gets arrow-key focus cycling for free. That would be the more complete fix, but it means a change to a different repository; this PR keeps the fix local to cosmic-files and scoped to the two delete-confirmation dialogs named in the issue.

Manual verification

Built and ran the debug binary directly (`target/debug/cosmic-files`, not the system-installed package) against a scratch folder with a test file:
1. Selected the file, pressed Shift+Delete to open "Excluir permanentemente?" — Delete is focused by default, as before.
2. Pressed Left/Up: focus ring moved to Cancel; pressed Right/Down: focus ring moved back to Delete. Confirmed both directions repeatedly.
3. Confirmed Enter activates whichever button currently has focus (Cancel dismisses the dialog, Delete deletes the file).
4. Confirmed the file grid behind the dialog no longer moves/reacts to arrow keys while the dialog is open (previously it did, since the key-bind loop ran unconditionally).
5. Confirmed Tab/Shift+Tab still work as before (unaffected — that path is untouched).

Repeated the same steps for the "Empty trash" permanently-delete dialog (`DialogPage::DeleteTrash`).

Automated tests

cargo build, cargo clippy and cargo test all pass with no new warnings (45 existing unit tests, unchanged). No new automated test was added: the behavior being fixed is keyboard-focus/redraw interaction with the widget runtime, which isn't exercised by this crate's existing (non-UI) unit test suite, so it's covered by the manual verification above instead.

Fixes #1990

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -8,8 +8,9 @@ use cosmic::iced::clipboard::dnd::DndAction;
 use cosmic::iced::core::SmolStr;
 use cosmic::iced::core::widget::operation::focusable::unfocus;
 use cosmic::iced::futures::{self, SinkExt};
-use cosmic::iced::keyboard::key::Physical;
+use cosmic::iced::keyboard::key::{Named, Physical};
 use cosmic::iced::keyboard::{Event as KeyEvent, Key, Modifiers};
+use cosmic::iced::widget::operation;
 #[cfg(all(feature = "wayland", feature = "desktop-applet"))]
 use cosmic::iced::platform_specific::shell::wayland::commands::overlap_notify::overlap_notify;
 use cosmic::iced::runtime::{clipboard, task};
@@ -601,6 +602,15 @@ pub enum DialogPage {
         path: PathBuf,
         entity: Entity,
     },
+}
+
+/// Returns true for dialogs whose two actions should support switching
+/// focus between them with arrow keys (in addition to Tab/Shift+Tab).
+fn dialog_supports_arrow_focus(dialog_page: &DialogPage) -> bool {
+    matches!(
+        dialog_page,
+        DialogPage::PermanentlyDelete { .. } | DialogPage::DeleteTrash { .. }
+    )
 }
 
 pub struct DialogPages {
@@ -3389,6 +3399,22 @@ impl Application for App {
                 #[cfg(not(all(feature = "wayland", feature = "desktop-applet")))]
                 let in_surface_ids = false;
                 if self.core.main_window_id() == Some(window_id) || in_surface_ids {
+                    if self
+                        .dialog_pages
+                        .front()
+                        .is_some_and(dialog_supports_arrow_focus)
+                    {
+                        match key {
+                            Key::Named(Named::ArrowRight) | Key::Named(Named::ArrowDown) => {
+                                return operation::focus_next();
+                            }
+                            Key::Named(Named::ArrowLeft) | Key::Named(Named::ArrowUp) => {
+                                return operation::focus_previous();
+                            }
+                            _ => {}
+                        }
+                    }
+
                     let entity = self.tab_model.active();
                     for (key_bind, action) in &self.key_binds {
                         if key_bind.matches(modifiers, &key, Some(&physical_key)) {


jacobgkau (changes requested):
The [PR template](https://github.com/pop-os/cosmic-files/blob/master/.github/PULL_REQUEST_TEMPLATE.md) that you removed from the description is mandatory. Please add it back, fill it in, and disclose any AI/LLM usage.

CarlosZambonii:
Updated: restored the PR template with all items checked, and added an explicit AI-disclosure line to the commit message (Claude Code was used to help implement and verify this fix; I reviewed and manually tested it myself before opening the PR). Force-pushed the amended commit.

PR #2015, opened 2026-08-26
fix: handle SFTP host key confirmation

### LINKED ISSUE #580: SFTP connection fails if host key not cached
**Cosmic-files version:** c7511bb

**Issue/Bug description:** When I try to connect to an SFTP share (SSH server) for the first time, I get an error dialog that says `Login dialog cancelled`.

If I connect to that IP address via `ssh` on the terminal, I get this normal message:

```
The authenticity of host '...' can't be established.
...
Are you sure you want to continue connecting (yes/no/[fingerprint])?
```

After accepting that prompt, which adds the host I'm connecting to to my local `~/.ssh/known_hosts` file, if I go back to `cosmic-files` and try to connect to SFTP again, I get the expected `Authentication Required` dialog.

**Steps to reproduce:** Connect to an SFTP server without having previously connected to that host via SSH.

**Expected behavior:** I should be prompted to accept the host's key fingerprint if it's the first time I'm connecting to that host.

**Other notes:** Nautilus handles this case.

alvinwin:
## Summary

GIO asks the mount operation to confirm an uncached SFTP host key, but COSMIC Files did not handle that question and reported `Login dialog cancelled`.

- forward generic mount-operation questions to the application
- display the backend-provided message and ordered choices
- return the selected choice without interpreting SSH policy
- prevent delayed dialog input from affecting a newer queued question

Fixes #580.

## Testing

- `cargo test --offline --all-features --lib --bins` — 45 passed
- `cargo check --offline --features gvfs`
- `git diff HEAD^ --check`
- first-connection SFTP verification against a disposable server with a fresh host key

Live verification on `c499cb9` was captured across two runs: one confirmed that Escape leaves the first-connection fingerprint question visible, and another confirmed that the same build can open the remote SFTP root. The combined Escape → continued mount sequence was not captured as one run. Automated coverage verifies that Escape preserves the active question and cannot dismiss the next queued question.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -672,6 +672,14 @@ impl DialogPages {
         matches!(self.pages.front(), Some(DialogPage::NetworkQuestion { .. }))
     }
 
+    fn pop_front_for_escape(&mut self) -> Option<(DialogPage, Task<Message>)> {
+        if self.front_is_network_question() {
+            None
+        } else {
+            self.pop_front()
+        }
+    }
+
     fn pop_network_question(
         &mut self,
         question_tx: &mpsc::Sender<i32>,
@@ -2825,8 +2833,11 @@ impl Application for App {
         let entity = self.tab_model.active();
 
         // Close dialog if open
-        if let Some((_page, task)) = self.dialog_pages.pop_front() {
-            return task;
+        if self.dialog_pages.front().is_some() {
+            return self
+                .dialog_pages
+                .pop_front_for_escape()
+                .map_or(Task::none(), |(_page, task)| task);
         }
 
         // Close gallery mode if open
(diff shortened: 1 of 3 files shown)


jacobgkau (commented):
This is the new dialog:

Screenshot_2026-09-01_13-32-26

I have some problems with the grammar of this message. However, it appears it's coming from GVFS, not COSMIC Files: https://github.com/GNOME/gvfs/blob/061802831d94e8aff45172f3ca18f953cf348b61/daemon/gvfsbackendsftp.c#L1350 So it may not be trivial for us to customize. And passing along GVFS's flawed message is objectively better than our current behavior of erroring out with no way to actually connect without resorting to other apps.

With that said, having both a `Cancel Login` and a `Cancel` button is awkward. I understand the former comes from GVFS and the latter comes from COSMIC Files, but users might not understand the difference. Does GVFS have any questions that wouldn't include a Cancel option?

It appears Nautilus, Thunar, and Nemo all pop the dialog with the expectation one of the options will be sufficient (and a lower-level option to abort is unnecessary):

Image

Image

Image

So we can probably just drop the second `Cancel` button and rely on GVFS to provide a cancel option in these questions, at least for now.

It would probably also be best to lay the buttons out horizontally, to better match our other dialogs. And if we can get the order flipped so the affirmative is farthest to the right (as those other three GVFS-using apps do), and add an appropriate icon, then that would also better match our other dialogs, e.g.:

Image

PR #2024, opened 2026-08-27
Allow reordering sidebar favorites by drag and drop

### LINKED ISSUE #309: [Feature Request] Drag to reorder in sidebar
Currently there doesn't appear to be any way to reorder the items in the sidebar. It would be great to be able to click and drag an element into a new spot.

cesarrew:
Fixes #309

Currently the only way to change the order of the favorites in the sidebar is to manually edit `~/.config/cosmic/com.system76.CosmicFiles/v1/favorites`. This PR makes the favorites in the sidebar reorderable by clicking and dragging them to a new spot, reusing the tab drag mechanism that libcosmic's `segmented_button` already provides (the same one cosmic-files already uses for reordering tabs in the tab bar).

## Changes

- The nav bar's inner segmented button is now built directly in `App::nav_bar` instead of going through the `nav_bar` widget wrapper, since the wrapper does not expose `enable_tab_drag`/`on_reorder`. The wrapper's styling (button metrics, scrollable, container style) is replicated so the sidebar looks exactly the same as before.
- A new `Message::ReorderNav` handler translates the widget's `ReorderEvent` into a new order for `config.favorites` and persists it, so the new order is saved and also picked up by other windows. Only favorites are reorderable; drops from/onto other sidebar entries (Recents, Trash, drives) are ignored, except that dropping a favorite just past the edges of the favorites section moves it to the start/end.
- `Message::DndEnterNav` now receives the offered mime types and ignores the reorder drag mime, so dragging a favorite over other sidebar entries no longer triggers the "navigate to hovered location" timeout meant for file drags. This mirrors what `DndEnterTab` already does for the tab drag mime.

Dropping files onto sidebar folders (copy/move) keeps working as before, as the reorder drag uses its own mime type (`x-cosmic-files/nav-dnd`).

## Testing

- `cargo test` (43 passed)
- `cargo check` with default features and with `--no-default-features --features wgpu` (no wayland)
- Ran the app and verified the sidebar renders identically to before

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01LM96Yx17hNGBtHNV58mv4F

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -35,7 +35,7 @@ use cosmic::widget::about::About;
 use cosmic::widget::dnd_destination::DragId;
 use cosmic::widget::menu::action::MenuAction;
 use cosmic::widget::menu::key_bind::KeyBind;
-use cosmic::widget::segmented_button::{self, Entity, ReorderEvent};
+use cosmic::widget::segmented_button::{self, Entity, InsertPosition, ReorderEvent};
 use cosmic::widget::{self, icon, settings, space};
 use cosmic::{Application, ApplicationExt, Element, cosmic_theme, executor, surface, theme};
 use mime_guess::Mime;
@@ -424,6 +424,7 @@ pub enum Message {
     PermanentlyDelete(Option<Entity>),
     Preview(Option<Entity>),
     ReloadMimeAppCache,
+    ReorderNav(ReorderEvent),
     ReorderTab(ReorderEvent),
     RescanRecents,
     RescanTrash,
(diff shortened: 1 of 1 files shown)


jacobgkau (changes requested):
The [PR template](https://github.com/pop-os/cosmic-files/blob/master/.github/PULL_REQUEST_TEMPLATE.md) that you removed from the description is mandatory. Please add it back and fill it in.

